PR #26, opened 2021-03-09
[CHE-245] iOS SDK Vault fixes 🏦

ceriksson:
CHE-245

# What this PR does

This PR updates the vault feature of the iOS SDK to fix some bugs and refactor part of the SDK.

This is part of the ongoing refactoring work for the SDK and aligns with upcoming changes to the official documentation.

# Before merging

_QA_

_Documentation_

- [x] If this PR modifies the API, I updated the API Reference.
- [x] If this PR adds new options to the API, I wrote some documentation or guide in the online documentation

# After merging


EvansPie:
I’m just leaving a couple of notes before I close the PR

`PrimerSessionFlow.default.vaulted` changed from `true` to `false` and default `uxMode` from `.vault` to `.checkout`. Is that the desired behaviour?

If that is what we need, then all the rest seem good 💯

ceriksson:
> I’m just leaving a couple of notes before I close the PR
> 
> `PrimerSessionFlow.default.vaulted` changed from `true` to `false` and default `uxMode` from `.vault` to `.checkout`. Is that the desired behaviour?
> 
> If that is what we need, then all the rest seem good 💯

Yeah, that's what we want :)

PR #28, opened 2021-03-11
Feature/che 251 abstract network layer

EvansPie:
CHE-251

# What this PR does

Abstract networking. You can find all the logic within the Services folder.

In two words:

- All API calls are defined by the `PrimerAPI` enum (which conforms to the `Endpoint` protocol)
- The `PrimerAPIClient` is responsible to actually making the request and returning a result.
- Result always is `.success(resultValue)` or `.failure`
- `PrimerAPIClient` will make the requests using its `networkService` (`URLSessionStack` object conforming into the `NetworkService` protocol)
- When you want to make an API request (e.g. `tokenize` a payment method) you need to construct a `PrimerAPIClient` object and call the corresponding function. Example below:

```
api.tokenizePaymentMethod(clientToken: clientToken, paymentMethodTokenizationRequest: request) { (result) in
            switch result {
            case .failure:
                onTokenizeSuccess(.failure( PrimerError.TokenizationRequestFailed ))
            case .success(let paymentMethodToken):
                onTokenizeSuccess(.success(paymentMethodToken))
            
            }
        }
```

Abstracting the network layer is the first step into decoupling logic into layers. I needed to refactor the logic a bit because the base URL is provided by the access token and is not explicitly defined.

Apart from replacing all API calls I didn't change any logic on success/failure.

Sorry beforehand for the huuuge commit 😅, but as soon as I removed the old networking logic I had to replace all the API calls.

# Before merging

_QA_

_Documentation_

# After merging


PrimerSDK/Services/API/Models/CreateAccountRequestParams.swift
```diff
@@ -0,0 +1,16 @@
+//
+//  CreateAccountRequest.swift
+//  primer-checkout-api
+//
+//  Created by Evangelos Pittas on 26/2/21.
+//
+
+import Foundation
+
+struct CreateAccountRequestParams {
```
ceriksson, line 10: τι είναι αυτό?

ceriksson (approved):
LGTM

We'll need to QA develop, but the code looks fine. Very clean 👍

PrimerSDK/Services/API/Models/CreateAccountRequestParams.swift
EvansPie: Nice catch!! Removed it, merged from develop and QA it. All seem to work.

PR #32, opened 2021-03-17
Feature/che 280 centralize error handling

EvansPie:
CHE-280

# What this PR does

No logic changed. Internal custom errors now conform on `CustomNSError` & `LocalizedError`. I.e. the carry with them an error domain, code, localized description and user info.

Added `AlertController` subclass that is actually constructing a new window on top of the current window and is being presented there. This avoids view hierarchy conflicts. Use as such `alert.show()`

Replaced all `print` console logs with `log`. `log` is a custom global function that doesn't print console logs on production builds, plus it provides log levels and applies styling on logs.

At the moment all API errors are passed through a centralized error handler that is just printing them. Further logic needs to be decided and applied in there.

When you print or show an error's localized description you should see the value returned by the error's enum `errorDescription`.

# Before merging

_QA_

_Documentation_

# After merging


PrimerSDK/Classes/Error Handler/Error.swift
```diff
@@ -0,0 +1,197 @@
+//
+//  Error.swift
+//  PrimerSDK
+//
+//  Created by Evangelos Pittas on 16/3/21.
+//
+
+import Foundation
+
+protocol PrimerErrorProtocol: CustomNSError, LocalizedError {
```
ceriksson, line 10: let's go through this file in a quick call too :)

PrimerSDK/Classes/Services/Network/URLSessionStack.swift
```diff
@@ -79,14 +79,56 @@ class URLSessionStack: NetworkService {
                 if let json = try? JSONSerialization.jsonObject(with: data, options: .allowFragments) as? [String: Any?],
                    let primerErrorJSON = json?["error"] as? [String: Any] {
                     let statusCode = (response as! HTTPURLResponse).statusCode
-                    let nsErr = NSError(domain: "Primer", code: 100, userInfo: primerErrorJSON)
                     
+                    let primerErrorResponse = try? self.parser.parse(PrimerErrorResponse.self, from: try! JSONSerialization.data(withJSONObject: primerErrorJSON, options: .fragmentsAllowed))
```
ceriksson, line 83: Let's go through this too :)

ceriksson (approved):
lgtm

PR #52, opened 2021-04-21
Feature/che 533 add payment method added to vault delegate method

EvansPie:
CHE-533

# What this PR does

Exposes a delegate function that returns the `PaymentMethodToken` when a payment method has been added in the vault.

Renames Klarna Customer Token to user's email.

Rename Saved Card to Saved Payment Method

I dropped the implementation of providing a callback to the developer, that he could use to notify the SDK that the token has been successfully sent to his/her backend. Reason is that it's not our responsibility if he/she sends the token successfully and Primer's job is just to tokenize and return the payment method. We shouldn't have any other responsibility. 

Moreover, if this were to be used to notify us that the integrating dev finished his call successfully and that we should show the success screen, this could come in conflict with showing/disabling the success screen from the settings on Primer init.

Finally, software design decisions should sync between the SDKs, so I leave that be finalized on the SDKs alignment task.

# Before merging

_QA_

_Documentation_

# After merging


ceriksson (approved):
lgtm 👍

PR #58, opened 2021-04-29
Feature/che 565 refactor flows

EvansPie:
CHE-565

# What this PR does

Refactors, removes and adds payment flows

None

# Before merging

_QA_

_Documentation_

# After merging

- 
Simulator Screen Shot - iPhone 12 - 2021-05-13 at 13 00 45
Simulator Screen Shot - iPhone 12 - 2021-05-13 at 13 00 30
Simulator Screen Shot - iPhone 12 - 2021-05-13 at 14 13 22


Sources/PrimerSDK/Classes/User Interface/Confirm Mandate/ConfirmMandateViewModel.swift
```diff
@@ -106,8 +106,10 @@ class ConfirmMandateViewModel: ConfirmMandateViewModelProtocol {
                     completion(error)
                 case .success(let token):
                     state.directDebitMandate = DirectDebitMandate(address: Address())
-                    let settings: PrimerSettingsProtocol = DependencyContainer.resolve()
-                    settings.onTokenizeSuccess(token, completion)
+                    // FIXME: Please review this carefully. I believe that the authorizePayment delegate method should not be called since Direct Debit can only be vaulted.
```
ceriksson, line 109: True, a direct debit token cannot be used to auth/capture a payment. However, it might still make sense to pass it back to the merchant as they might want to store the token or something? Wdyt?

ceriksson (commented):
Looks good, let's align on comments then merge :)

Sources/PrimerSDK/Classes/Data Models/PrimerSessionFlow.swift
```diff
@@ -36,58 +36,59 @@
...
 
-    case `default`
-    case defaultWithVault
-    case completeDirectCheckout
-    case addPayPalToVault
-    case addCardToVault
-    case addDirectDebitToVault
-    case addKlarnaToVault
-    case addDirectDebit
+    case primerWithVault
```
xevious78, line 39: [Feedback]
We can simplify the naming of `primerWithVault` and `primerCheckout` - having `primer` as part of the name doesn't give that much information about what this does.
I think we can just call `vault` and `checkout`.

xevious78 (changes requested):
Good first step but I think we should go further 💪 

This is the kind of UI the developer shouldn't be able to create: the user is asked to *vault* a payment method for a *payment*. 
CleanShot 2021-05-03 at 18 36 27@2x

From what I see, there's a bit of confusion between the "vault" flow (using the checkout to vault a payment method) and the "vault" manager (the UI that allows quick payment). 
What I suggest is:
- `primerWithVault` renamed `vault` that enables the user to vault a payment method among a list of payment methods. The dev shouldn't be able to show the vault manager & the amount.
- `primerCheckout` renamed `checkout` that enables the user to pay with a payment method, including a vaulted one. This is the Universal Checkout. Hiding the vault manager should be a setting of the flow, and not a flow in and on itself.

@EvansPie @ceriksson what do you think?

Sources/PrimerSDK/Classes/Data Models/PrimerSessionFlow.swift
EvansPie: Done

Sources/PrimerSDK/Classes/Data Models/PrimerSessionFlow.swift
```diff
@@ -36,58 +36,59 @@
...
+    case checkoutWithPayPal
+    case vaultDirectDebit
+    case vaultKlarna
     case checkoutWithKlarna
-
+    
     var vaulted: Bool {
         switch self {
-        case .addCardToVault:
+        case .primerWithVault:
```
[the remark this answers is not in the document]
EvansPie: Done

Sources/PrimerSDK/Classes/User Interface/Confirm Mandate/ConfirmMandateViewModel.swift
EvansPie: Makes sense, fixed!

Sources/PrimerSDK/Classes/Data Models/PrimerSessionFlow.swift
EvansPie: @xevious78 I agree on what you say above. I believe the above screenshot is an outcome of not checking the flow and just adding the amount if the developer has provided one. We have to change this. I'll create a ticket for this one.

xevious78: @EvansPie I can pick this up and make sure that the flows are properly represented. Let's have a quick sync on this :)

ceriksson (commented):
Will let Aladin approve this once changes have been made :)

Sources/PrimerSDK/Classes/Data Models/PrimerInternalSessionFlow.swift
```diff
@@ -46,49 +45,72 @@ public enum PrimerSessionFlow {
...
+            return .vaultKlarna
+        case .addDirectDebit:
+            return .vaultDirectDebit
+        case .checkoutWithKlarna:
+            return .checkoutWithKlarna
+        }
+    }
+}
+
+internal enum PrimerInternalSessionFlow {
```
ceriksson, line 72: is this so we dont expose the singleton to the developer?

ceriksson (commented):
lgtm, just waiting for the screenshots, then I'll approve

Sources/PrimerSDK/Classes/Data Models/PrimerInternalSessionFlow.swift
EvansPie: This isn't a singleton, just an enum. This is so we can use naming that makes more sense internally, while not making breaking changes.

ceriksson (approved):
lgtm 👍

xevious78 (approved):

PR #59, opened 2021-04-29
Feature/che 475 update labels

EvansPie:
CHE-475

# What this PR does

Update labels' texts according to Figma

There were a few labels on Figma that didn't exist on our design. I tried modifying the UI but right now it's too complicated to do something about it and I'm really worried we'll break tons of stuff. I'm gonna come up with a better solution for this.

# Before merging

_QA_

_Documentation_

# After merging


Sources/PrimerSDK/Classes/Data Models/PrimerContent.swift
```diff
@@ -110,8 +110,8 @@ struct VaultPaymentMethodViewContent {
     var mainTitleText: String { return NSLocalizedString("primer-vault-payment-method-saved-payment-methods",
                                                          tableName: nil,
                                                          bundle: Bundle.primerResources,
-                                                         value: "Saved payment methods",
-                                                         comment: "Saved payment methods - Vault Payment Method (Main title text)") }
+                                                         value: "SAVED PAYMENT METHODS",
```
xevious78, line 113: [Feedback]
The fact that it is all uppercase is a UI decision, independent from the text itself. Can we keep lowercase here but use a function to put the string to uppercase when needed?
Same thing would go for `OTHER WAYS TO PAY`

xevious78 (changes requested):
🚀 
@EvansPie : you told me in a call that you couldn't add some labels. Is that still the case?

I don't see `AVAILABLE PAYMENT METHODS` that is displayed on the Universal Checkout when the Vault Manager is not visible 🤔

Sources/PrimerSDK/Classes/Data Models/PrimerContent.swift
EvansPie: Fixed

PR #61, opened 2021-05-05
Feature/che 622 refactor token completion

EvansPie:
CHE-622

# What this PR does

Replace CreateClientTokenResponse to String for the client token

n/a

# Before merging

_QA_

_Documentation_

# After merging


Sources/PrimerSDK/Classes/Data Models/PrimerSettings.swift
```diff
@@ -1,6 +1,6 @@
 #if canImport(UIKit)
 
-public typealias ClientTokenCallBack = (_ completionHandler: @escaping (Result<CreateClientTokenResponse, Error>) -> Void) -> Void
+public typealias ClientTokenCallBack = (_ completionHandler: @escaping (Result<String, Error>) -> Void) -> Void
```
ceriksson, line 3: down the line, maybe we wrap this in a struct something? (keeping in mind the feedback regarding decodable etc.). Seems like an easy way to introduce bugs passing an anonymous strings around.

Sources/PrimerSDK/Classes/Extensions & Utilities/StringExtension.swift
```diff
@@ -82,25 +82,16 @@ extension String {
...
-                    } catch {
-                        log(logLevel: .error, title: "PARSER", message: "Failed to parse client token", prefix: nil, suffix: nil, bundle: Bundle.primerFrameworkIdentifier, file: #file, className: nil, function: #function, line: #line)
-                    }
-                }
-            }
-
-        }
-        return DecodedClientToken()
+    
+    var decodedJWTToken: DecodedClientToken? {
```
ceriksson, line 86: is this tested?

ceriksson (approved):
looks fine, may wanna double check the new base 64 decoding works properly.

Sources/PrimerSDK/Classes/Data Models/PrimerSettings.swift
EvansPie: Well, it's a very usual practice to pass primitive vars if possible. This is what Mayur wanted, to not have to create a custom decodable object just as a string wrapper. Moreover, now we don't even have the expiration timestamp in this object, so it'd be a struct containing just a string.

Sources/PrimerSDK/Classes/Extensions & Utilities/StringExtension.swift
EvansPie: Yeah, quite a lot actually. It's one of the snippets that I was using on my previous company. I'd need to write some test cases though. I do think however that I should change the name, cause this isn't the whole token information, just the payload. @xevious78 @ceriksson would we need any other information from the token? Maybe I should ask Will about it

PR #72, opened 2021-05-24
Feature/che 408 apple pay

EvansPie:
CHE-408

# What this PR does

Delay dismissing apple pay view, show success/error screen.

n/a

# Before merging

_QA_

_Documentation_

# After merging


Sources/PrimerSDK/Classes/Data Models/PaymentNetwork.swift
```diff
@@ -0,0 +1,154 @@
...
+import Foundation
+import PassKit
+
+public enum PaymentNetwork: String {
+    
+    case chinaUnionPay
+    
+    case discover
+    
+    @available(iOS 12.0, *)
```
xevious78, line 19: [Feedback]
These payment networks will also be used in the card form (to filter out specific card networks), and they should be available regardless of the version of iOS.
Not sure we should restrict the availability of these payment networks. Unless you plan to have a different data models for card networks specifically?

Sources/PrimerSDK/Classes/Error Handler/Error.swift
```diff
@@ -114,6 +115,13 @@ enum KlarnaException: PrimerErrorProtocol {
                                      bundle: Bundle.primerResources,
                                      value: "Failed to find Klarna order items",
                                      comment: "Failed to find Klarna order items - Error message")
+            
+        case .orderItemMissesAmount:
+            return NSLocalizedString("primer-klarna-error-message-missing-order-item-amount",
+                                     tableName: nil,
+                                     bundle: Bundle.primerResources,
+                                     value: "Failed to find amount for an order item",
```
xevious78, line 123: [Q]
Is there any way this message gets displayed to the user? A user can't do anything with this 😅

xevious78: Same questions for the pending true below

Sources/PrimerSDK/Classes/Core/Payment Services/ApplePayService.swift
```diff
@@ -0,0 +1,84 @@
...
+}
+
+enum ApplePayType {
+    case recurring, checkout
+}
+
+class ApplePayService: NSObject, ApplePayServiceProtocol {
+    
+    deinit {
+        print("ApplePayService deinit")
```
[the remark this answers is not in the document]
xevious78: Would it be better to have `#if DEBUG` here?

Sources/PrimerSDK/Classes/User Interface/Apple Pay/ApplePayViewController+PassKit.swift
```diff
@@ -1,112 +1,113 @@
-#if canImport(UIKit)
```
xevious78, line 1: [Q]
What's the plan with this file?

Sources/PrimerSDK/Classes/User Interface/Apple Pay/ApplePayViewController.swift
```diff
@@ -1,38 +1,38 @@
-#if canImport(UIKit)
```
xevious78, line 1: [Q]
What's the plan with this file?

Sources/PrimerSDK/Classes/User Interface/Apple Pay/ApplePayViewModel.swift
```diff
@@ -1,69 +1,317 @@
...
+        let applePayRequest = ApplePayRequest(
+            currency: currency,
+            merchantIdentifier: merchantIdentifier,
+            countryCode: countryCode,
+//            supportedNetworks: supportedNetworks,
+            items: orderItems
+//            merchantCapabilities: merchantCapabilities
+        )
+        
+        var supportedNetworks: [PKPaymentNetwork] = [
```
xevious78, line 84: [Suggestion]
Would be great to have this whole logic of supported networks on another file. It's making reading the code a bit tedious 👍

Sources/PrimerSDK/Classes/User Interface/Apple Pay/ApplePayViewModel.swift
```diff
@@ -1,69 +1,317 @@
...
+                case .success(let applePayPaymentResponse):
+                    // Loading view
+                    let settings: PrimerSettingsProtocol = DependencyContainer.resolve()
+                    
+                    let applePayService: ApplePayServiceProtocol = DependencyContainer.resolve()
+                    applePayService.fetchConfig { [weak self] (err) in
+                        if let err = err {
+                            DispatchQueue.main.async {
+                                Primer.shared.presentingViewController?.dismiss(animated: true, completion: {
+                                    if !settings.hasDisabledSuccessScreen {
```
xevious78, line 168: [Feedback]
Apple Pay shouldn't care about what's after 🤔 
It could be a success screen or it could be something else in the future. Is it possible to reduce the coupling between Apple Pay and the bottom sheet?

Sources/PrimerSDK/Classes/User Interface/Apple Pay/ApplePayViewModel.swift
```diff
@@ -1,69 +1,317 @@
...
+                                case .success(let token):
+                                    DispatchQueue.main.async {
+                                        if Primer.shared.flow.internalSessionFlow.vaulted {
+                                            Primer.shared.delegate?.tokenAddedToVault(token)
+                                            completion(nil)
+                                        } else {
+                                            //settings.onTokenizeSuccess(token, completion)
+                                            Primer.shared.delegate?.authorizePayment(token, { (err) in
+                                                DispatchQueue.main.async {
+                                                    Primer.shared.presentingViewController?.dismiss(animated: true, completion: {
```
xevious78, line 224: Callback heeeeeell! :D

Sources/PrimerSDK/Classes/User Interface/Apple Pay/ApplePayViewModel.swift
```diff
@@ -1,69 +1,317 @@
...
+                                                        if !settings.hasDisabledSuccessScreen {
+                                                            Primer.shared.root = RootViewController()
```
xevious78, line 226: [Feedback]
Seems that you wrote this piece of code (to go back to the bottom sheet and display the success/error screen) three times. Could we abstract this a bit more?

Sources/PrimerSDK/Classes/User Interface/Apple Pay/ApplePayViewModel.swift
```diff
@@ -1,69 +1,317 @@
...
+            applePayCompletion = nil
+        } catch {
+            applePayCompletion?(.failure(error))
+            applePayCompletion = nil
+        }
+    }
+    
+}
+
+extension PKPaymentMethodType {
```
xevious78, line 300: [Q]
Is it a good iOS practice to extend these types?

Sources/PrimerSDK/Classes/User Interface/Checkout/VaultCheckoutViewController.swift
```diff
@@ -103,7 +103,18 @@ extension VaultCheckoutViewController: UITableViewDelegate, UITableViewDataSourc
...
         case .applePay:
-            router.show(.applePay)
+            let router: RouterDelegate = DependencyContainer.resolve()
+            let rvc = router.root
+            rvc?.dismiss(animated: true, completion: {
+                DispatchQueue.main.async { [weak self] in
+                    let appleViewModel: ApplePayViewModelProtocol = DependencyContainer.resolve()
+                    appleViewModel.payWithApple { (err) in
+                        if let err = err {
+                            print(err)
```
xevious78, line 113: [Feedback]
Doesn't look like a good error management to just print an error in the logs 😅

Sources/PrimerSDK/Classes/User Interface/Apple Pay/ApplePayViewModel.swift
```diff
@@ -1,69 +1,317 @@
...
+//            @available(iOS 14.0, *)
+//            supportedNetworks.append(.barcode)
+            supportedNetworks.append(.girocard)
+        }
+        
+        if PKPaymentAuthorizationViewController.canMakePayments(usingNetworks: supportedNetworks) {
+            let request = PKPaymentRequest()
+            request.currencyCode = applePayRequest.currency.rawValue
+            request.countryCode = applePayRequest.countryCode.rawValue
+            request.merchantIdentifier = "merchant.primer.dev.evangelos"
```
xevious78, line 145: [Feedback]
This shouldn't end up in the final code. I think the merchant identifier should come from the configuration?

Sources/PrimerSDK/Classes/Core/Payment Services/ApplePayService.swift
```diff
@@ -0,0 +1,84 @@
...
+
+                state.viewModels = []
+
+                config.paymentMethods?.forEach({ method in
+                    guard let type = method.type else { return }
+                    if type == .googlePay { return }
+                    state.viewModels.append(PaymentMethodViewModel(type: type))
+                })
+
+                // ensure Apple Pay is always first if present.
```
xevious78, line 52: [Q]
Why do you want to ensure this?

Sources/PrimerSDK/Classes/Core/Payment Services/ApplePayService.swift
```diff
@@ -0,0 +1,84 @@
...
+//
+//  Created by Evangelos Pittas on 16/4/21.
+//
+
+import Foundation
+
+
+protocol ApplePayServiceProtocol {
+    func tokenize(instrument: PaymentInstrument, completion: @escaping (Result<PaymentMethodToken, Error>) -> Void)
+    func fetchConfig(_ completion: @escaping (Error?) -> Void)
```
xevious78, line 13: [Q]
This function is called `fetchConfiguration` at the API level. Which convention you want to use? :)

xevious78 (changes requested):
Nice work :) Got some comments!

Sources/PrimerSDK/Classes/Core/Payment Services/ApplePayService.swift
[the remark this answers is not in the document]
EvansPie: Yep, nice catch. Forgot to remove it. Fixed.

Sources/PrimerSDK/Classes/Core/Primer/Primer.swift
```diff
@@ -152,19 +155,28 @@ public class Primer {
...
-            guard let root = self?.root else { return }
-            let router: RouterDelegate = DependencyContainer.resolve()
-            let settings: PrimerSettingsProtocol = DependencyContainer.resolve()
-            
-            if flow.internalSessionFlow.vaulted {
-                (settings as! PrimerSettings).amount = nil
+            if case .checkoutWithApplePay = flow {
+                let appleViewModel: ApplePayViewModelProtocol = DependencyContainer.resolve()
+                appleViewModel.payWithApple { (err) in
+
```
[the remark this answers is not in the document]
EvansPie: Nope, cause the error management is performed (and exposed) in the view model. The major difference between Apple Pay and other services, is that Apple Pay doesn't present the root view controller (as the other services do). So on the others, the root view controller is acting as a proxy. Also, because the root view controller is responsible for showing success/error screen, it was made responsible to expose the errors. I don't agree with this, but it's something that will take quite some effort to be refactored, and I consider it out of scope of this ticket.

Sources/PrimerSDK/Classes/Core/Primer/Primer.swift
```diff
@@ -1,5 +1,6 @@
 #if canImport(UIKit)
 
+import PassKit
```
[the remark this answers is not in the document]
EvansPie: You're right, I removed it.

Sources/PrimerSDK/Classes/Data Models/OrderItem.swift
```diff
@@ -6,19 +6,38 @@
...
 
 public struct OrderItem: Codable {
     public let name: String
-    public let unitAmount: Int
+    public let unitAmount: Int?
     public let quantity: Int
+    public var isPending: Bool = false
+    
+    public var applePayItem: PKPaymentSummaryItem {
+        let item = PKPaymentSummaryItem(label: name, amount: NSDecimalNumber(value: Double(unitAmount ?? 0)/100))
```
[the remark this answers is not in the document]
EvansPie: Got it. Let's create a separate ticket for that.

Sources/PrimerSDK/Classes/User Interface/Apple Pay/ApplePayViewModel.swift
EvansPie: Well, it's very common when you want to add functionality on existing classes. This is not the same case as with the conflicts that Voi encountered, since `PKPaymentMethodType` cannot be subclassed. I just modified the access level to internal.

Also, since I've chosen a name that starts with `primer`, I consider it quite difficult another developer to want to have the same extension name.

Sources/PrimerSDK/Classes/User Interface/Checkout/VaultCheckoutViewController.swift
EvansPie: Same case as above comment, but from the universal checkout entry point. 

https://github.com/primer-io/primer-sdk-ios/pull/72#discussion_r639692573

Sources/PrimerSDK/Classes/User Interface/Apple Pay/ApplePayViewModel.swift
EvansPie: 🤫
Fixed

Sources/PrimerSDK/Classes/Core/Payment Services/ApplePayService.swift
EvansPie: This was there in the code before I join. I thought it's some kind of business decision that I wasn't aware of. @ceriksson may have an insight. Do you want me to remove it?

Sources/PrimerSDK/Classes/Core/Payment Services/ApplePayService.swift
EvansPie: Well.. you're right. But the other services have it like `fetchConfig`, so I kept the one from the other services. To tell you the truth, I can rename all of those, they are internal and won't break anything for anyone.

Example/PrimerSDK/MerchantCheckoutViewController.swift
```diff
@@ -33,6 +33,25 @@ class MerchantCheckoutViewController: UIViewController {
         isInitialLoadingHidden: true
     )
     
+    let applePaySettings = PrimerSettings(
+        merchantIdentifier: "merchant.primer.dev.evangelos",
+        currency: .EUR,
+        countryCode: .fr,
+        applePayEnabled: true,
```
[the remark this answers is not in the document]
xevious78: Nobody is using Apple Pay, it think it's safe to remove now and not accumulate tech debt :)

Sources/PrimerSDK/Classes/User Interface/Checkout/VaultCheckoutViewController.swift
EvansPie: Removed it completely. That was only for me.

Sources/PrimerSDK/Classes/Core/Payment Services/ApplePayService.swift
ceriksson: this was a design request, something with Apple wanting Apple Pay to be displayed as the first option. Not sure whether this is a crucial requirement by Apple. If we want to I believe we can leave it out :)

xevious78: Oh yeah I remember. We can keep it like this but let's have a proper comment + something on the documentation about this info :)

Sources/PrimerSDK/Classes/Core/Payment Services/ApplePayService.swift
xevious78: I leave it up to you to decide :) maybe we can have a separate PR for this internal renaming?

Sources/PrimerSDK/Classes/Core/Payment Services/ApplePayService.swift
EvansPie: Cool, will add a comment

Sources/PrimerSDK/Classes/Core/Payment Services/ApplePayService.swift
EvansPie: Yeah, actually I'm going to do it alongside with the other rename.

Sources/PrimerSDK/Classes/Data Models/PaymentNetwork.swift
EvansPie: At the moment this is used only by Apple Pay. You're right though, a more generic model is needed since we can re-use this on the card models.

Fixed.

Sources/PrimerSDK/Classes/Error Handler/Error.swift
EvansPie: No, this error is the error thrown to the developer. Currently there's no alert showing from the SDK at all, we removed them a couple of weeks back.

Sources/PrimerSDK/Classes/User Interface/Apple Pay/ApplePayViewController+PassKit.swift
EvansPie: No plan actually, this was Carl's implementation that I commented out so I could be having a look on specific decisions that he wanted to take. Removing it.

Sources/PrimerSDK/Classes/User Interface/Apple Pay/ApplePayViewController.swift
EvansPie: Same as above

Sources/PrimerSDK/Classes/User Interface/Checkout/VaultCheckoutViewController.swift
```diff
@@ -103,7 +103,16 @@ extension VaultCheckoutViewController: UITableViewDelegate, UITableViewDataSourc
 
         switch option.type {
         case .applePay:
-            router.show(.applePay)
+            let router: RouterDelegate = DependencyContainer.resolve()
+            let rvc = router.root
+            rvc?.dismiss(animated: true, completion: {
```
ceriksson, line 108: rvc = rootViewController?

Sources/PrimerSDK/Classes/User Interface/Root/RootViewController+Router.swift
```diff
@@ -10,6 +10,7 @@
 import UIKit
 
 protocol RouterDelegate: class {
+    var root: RootViewController? { get }
```
ceriksson, line 13: following my comment above, should we call it root or rvc or rootViewController?

ceriksson (approved):
nice! left some minor questions

Sources/PrimerSDK/Classes/User Interface/Apple Pay/ApplePayViewModel.swift
EvansPie: Makes sense. Done!

Sources/PrimerSDK/Classes/User Interface/Apple Pay/ApplePayViewModel.swift
EvansPie: Same as here https://github.com/primer-io/primer-sdk-ios/pull/72#discussion_r639692573

Sources/PrimerSDK/Classes/User Interface/Apple Pay/ApplePayViewModel.swift
EvansPie: Yep :( 

We do need to assign some time to embed PromiseKit

Sources/PrimerSDK/Classes/User Interface/Apple Pay/ApplePayViewModel.swift
EvansPie: Moved into the Router.

Sources/PrimerSDK/Classes/User Interface/Root/RootViewController+Router.swift
EvansPie: Leave it root, that's fine. On the later version let's call it rvc.

Sources/PrimerSDK/Classes/Error Handler/Error.swift
xevious78: @ceriksson: this type of error shouldn't be given to any user, as it doesn't make any sense for them.
@EvansPie: since this error is and will only be for developers,  can we remove it from the localised string?

EvansPie: It's common on iOS errors to be localized. They do have a property `localizedDescritpion`, which gives you the error description.

xevious78: Fair enough. It's a bit weird to me to take some time to translate developer error x) Let's keep it as-is for now

xevious78 (approved):

PR #77, opened 2021-05-31
Fastlane release on Appetize + Firebase

xevious78:


ceriksson (approved):
nice with the comments 👍

PR #81, opened 2021-06-03
Feature/che 781 UI fixes

EvansPie:
CHE-781

# What this PR does

- Fixes button text clipping.
- Change Apple Pay color to black (Apple's default). Discuss if we want this customizable.
- Fix vault manager wrong subtitle copy.
- Fix bug that amount was nullified and amount was not showing on universal checkout.

n/a

# Before merging

_QA_

_Documentation_

# After merging


Sources/PrimerSDK/Classes/Core/Primer/Primer.swift
```diff
@@ -167,12 +167,6 @@ public class Primer {
                 self?.root = RootViewController()
                 guard let root = self?.root else { return }
                 let router: RouterDelegate = DependencyContainer.resolve()
-                let settings: PrimerSettingsProtocol = DependencyContainer.resolve()
-                
-                if flow.internalSessionFlow.vaulted {
```
ceriksson, line 172: What's happening here?

ceriksson (commented):
👍 just got some questions regarding dependencies removed + test coverage

Sources/PrimerSDK/Classes/Core/Primer/Primer.swift
EvansPie: Those lines were nullifying the amount in the vault flow, leading to a bug if you opened the vault flow and then the checkout flow. We don't need to nullify the amount (as it is one of the variables configured by the dev), we need to not present it in the vault flow.

ceriksson: makes sense 👍

Sources/PrimerSDK/Classes/User Interface/Apple Pay/ApplePayViewModel.swift
```diff
@@ -57,19 +57,31 @@ class ApplePayViewModel: NSObject, ApplePayViewModelProtocol {
     // swiftlint:disable cyclomatic_complexity function_body_length
     func payWithApple(completion: @escaping (Error?) -> Void) {
         guard let countryCode = countryCode else {
-            return completion(PaymentException.missingCountryCode)
+            let err = PaymentException.missingCountryCode
```
xevious78, line 60: [Suggestion]
Loads of duplicated code here. Would it be possible to put the common stuff in some function?

Sources/PrimerSDK/Classes/User Interface/Checkout/VaultCheckoutViewController.swift
```diff
@@ -107,7 +107,7 @@ extension VaultCheckoutViewController: UITableViewDelegate, UITableViewDataSourc
             router.root?.dismiss(animated: true, completion: {
                 DispatchQueue.main.async { [weak self] in
                     let appleViewModel: ApplePayViewModelProtocol = DependencyContainer.resolve()
-                    appleViewModel.payWithApple { (_) in
+                    appleViewModel.payWithApple { (err) in
```
xevious78, line 110: [Q]
We don't do anything with the error? What's the point of this change then?

xevious78 (approved):
In terms of fixes: it does the job! Just got a couple of non-blocking questions/comments :)

Sources/PrimerSDK/Classes/User Interface/Vault/VaultPaymentMethodViewController.swift
```diff
@@ -22,9 +20,169 @@ internal class VaultPaymentMethodViewController: PrimerViewController {
...
+    }
+
+    func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        let viewModel: VaultPaymentMethodViewModelProtocol = DependencyContainer.resolve()
+        
+        if indexPath.row == viewModel.paymentMethods.count {
+            return
+        }
+
+        tableView.deselectRow(at: indexPath, animated: true)
```
xevious78, line 67: [Q]
Why do you need to deselect the row when it has been selected?

PR #82, opened 2021-06-03
Ship to firebase on preview

xevious78:
This PR ships to firebase when a PR is created OR a PR is updated with a new commit


ceriksson (approved):
nice!

PR #86, opened 2021-06-15
[CHE-798] Show error message for failed PayPal init

ceriksson:
CHE-798

# What this PR does

When the developer has not disabled error UI, the error view is shown on error.

The scroll view acts weird in the simulator. Let's discuss below.

Also, let's discuss ways to write a UI test for this scenario. It's been tested on live device for both PayPal & Klarna but testing requires tinkering with the example app config files we pass in on init.

_QA_

_Documentation_

# After merging

### DIFF
--- a/Sources/PrimerSDK/Classes/User Interface/OAuth/OAuthViewController.swift
+++ b/Sources/PrimerSDK/Classes/User Interface/OAuth/OAuthViewController.swift
@@ -49,9 +49,12 @@ internal class OAuthViewController: PrimerViewController {
                     
                     let settings: PrimerSettingsProtocol = DependencyContainer.resolve()
                     
+                    let routerDelegate: RouterDelegate = DependencyContainer.resolve()
+                    
+                    let router = routerDelegate as! Router
+                    
                     if settings.hasDisabledSuccessScreen {
-                        let routerDelegate: RouterDelegate = DependencyContainer.resolve()
-                        let router = routerDelegate as! Router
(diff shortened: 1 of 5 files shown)


Sources/PrimerSDK/Classes/User Interface/OAuth/OAuthViewController.swift
```diff
@@ -49,16 +49,21 @@ internal class OAuthViewController: PrimerViewController {
...
-                        let router = routerDelegate as! Router
+                        
                         let rootViewController = router.root
                         
                         UIView.animate(withDuration: 0.3) {
                             (rootViewController?.presentationController as? PresentationController)?.blurEffectView.alpha = 0.0
                         } completion: { (_) in
                             rootViewController?.dismiss(animated: true, completion: nil)
                         }
+                    } else {
```
EvansPie, line 65: I don't remember that clearly, but I think there's a reason I hadn't added this 'else' statement. Have you checked that it doesn't break the Klarna flow, cause OAuthViewController is re-used.

ceriksson: Cheers yeah, I'll be performing some more QA, especially for Klarna (just wanted to start the discussion here). Still, it's probably best if we enable this UI for PayPal at least. MIght refactor and do this only for PayPal.

EvansPie (approved):

PR #88, opened 2021-06-15
Feature/che 723 klarna deep linking redirection

EvansPie:
CHE-723

# What this PR does

Listen webview redirects and open bankid app if it exists.

n/a

# Before merging

_QA_

_Documentation_

# After merging

### DIFF
--- a/Example/PrimerSDK/Info.plist
+++ b/Example/PrimerSDK/Info.plist
@@ -56,5 +56,9 @@
 	<array>
 		<string>UIInterfaceOrientationPortrait</string>
 	</array>
+    <key>LSApplicationQueriesSchemes</key>
+    <array>
+        <string>bankid</string>
+    </array>
 </dict>
 </plist>
--- a/Sources/PrimerSDK/Classes/User Interface/OAuth/WebViewController.swift
+++ b/Sources/PrimerSDK/Classes/User Interface/OAuth/WebViewController.swift
@@ -57,6 +57,17 @@ internal class WebViewController: PrimerViewController, WKNavigationDelegate {
     func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, decisionHandler: @escaping (WKNavigationActionPolicy) -> Void) {
         log(logLevel: .info, message: "🚀 \(navigationAction.request.url?.host ?? "n/a")")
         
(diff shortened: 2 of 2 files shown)


Example/PrimerSDK/Info.plist
```diff
@@ -56,5 +56,9 @@
 	<array>
 		<string>UIInterfaceOrientationPortrait</string>
 	</array>
+    <key>LSApplicationQueriesSchemes</key>
+    <array>
+        <string>bankid</string>
```
ceriksson, line 61: does this mean every merchant needs to add this to their app?

Sources/PrimerSDK/Classes/User Interface/OAuth/WebViewController.swift
```diff
@@ -57,6 +57,21 @@ internal class WebViewController: PrimerViewController, WKNavigationDelegate {
     func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, decisionHandler: @escaping (WKNavigationActionPolicy) -> Void) {
         log(logLevel: .info, message: "🚀 \(navigationAction.request.url?.host ?? "n/a")")
         
+        if var urlStr = navigationAction.request.url?.absoluteString, urlStr.hasPrefix("bankid://") == true {
+            let settings: PrimerSettingsProtocol = DependencyContainer.resolve()
+            if urlStr.contains("redirect=null"), let urlScheme = settings.urlScheme {
+                urlStr = urlStr.replacingOccurrences(of: "redirect=null", with: "redirect=\(urlScheme)")
```
ceriksson, line 63: we may need to ascertain with BankID & Klarna that this will work in the future, or at least make sure we have a back up in case it doesn't work.

ceriksson (commented):
Just some comments so we align. I will go ahead and do the QA now :)

Example/PrimerSDK/Info.plist
EvansPie: yeah

Sources/PrimerSDK/Classes/User Interface/OAuth/WebViewController.swift
ceriksson: cool sounds good. So long as we don't break anything.

ceriksson (approved):
Happy now, let's double check the existing Klarna flows works as expected then let's merge this.

PR #97, opened 2021-06-24
Improvement/make unnecessary methods in primer delegate optional

mayur-ios-dev:
CHE-

# What this PR does

Describe ... (Features / UI / logic changes)

List & explain ...

# Before merging

_QA_

_Documentation_

# After merging

### DIFF
--- a/Sources/PrimerSDK/Classes/Core/Primer/PrimerDelegate.swift
+++ b/Sources/PrimerSDK/Classes/Core/Primer/PrimerDelegate.swift
@@ -34,6 +34,13 @@ public protocol PrimerDelegate: class {
     func checkoutFailed(with error: Error)
 }
 
+public extension PrimerDelegate {
+    func tokenAddedToVault(_ token: PaymentMethodToken) {}
+    func authorizePayment(_ result: PaymentMethodToken, _ completion:  @escaping (Error?) -> Void) {}
+    func onCheckoutDismissed() {}
+    func checkoutFailed(with error: Error) {}
+}
+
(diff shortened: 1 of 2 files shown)


Sources/PrimerSDK/Classes/Core/Primer/PrimerDelegate.swift
```diff
@@ -34,6 +34,13 @@ public protocol PrimerDelegate: class {
     func checkoutFailed(with error: Error)
 }
 
+public extension PrimerDelegate {
```
EvansPie, line 37: lgtm! 

Thing is that this is the first thing I tried, and I remember getting an error. I have no idea why, and I don't remember what it was.

EvansPie (commented):
The only problem I noticed is the one in the screenshot, that we'll not be able to mark functions as unavailable in the future. I do prefer this solution though (shorter & cleaner), rather than adding `@objc` flag on the protocol.

I'll have to discuss this with Aladin and take a decision.

Screenshot 2021-06-25 at 2 05 31 PM

mayur-ios-dev:
> The only problem I noticed is the one in the screenshot, that we'll not be able to mark functions as unavailable in the future. I do prefer this solution though (shorter & cleaner), rather than adding `@objc` flag on the protocol.
> 
> I'll have to discuss this with Aladin and take a decision.
> 
> Screenshot 2021-06-25 at 2 05 31 PM

I think we should be able to make them unavailable by using 

```swift
#if {Condition for availability}
// declaration here
#endif
```

PR #99, opened 2021-07-06
Feature/che 896 pay pal checkout fix

EvansPie:
CHE-896

# What this PR does

Fixes bug on PayPal checkout, app was staling.

n/a

# Before merging

_QA_

_Documentation_

# After merging

### DIFF
--- a/Sources/PrimerSDK/Classes/Core/Payment Services/PayPalService.swift
+++ b/Sources/PrimerSDK/Classes/Core/Payment Services/PayPalService.swift
@@ -57,16 +57,20 @@ internal class PayPalService: PayPalServiceProtocol {
             return completion(.failure(PrimerError.payPalSessionFailed))
         }
 
-        guard let urlScheme = settings.urlScheme else {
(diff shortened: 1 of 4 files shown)


Sources/PrimerSDK/Classes/Core/Payment Services/PayPalService.swift
```diff
@@ -57,16 +57,20 @@ internal class PayPalService: PayPalServiceProtocol {
...
+            urlScheme = urlScheme.replacingOccurrences(of: "://", with: "")
         }
 
         let body = PayPalCreateOrderRequest(
             paymentMethodConfigId: configId,
             amount: amount,
             currencyCode: currency,
-            returnUrl: urlScheme,
-            cancelUrl: urlScheme
+            returnUrl: "\(urlScheme)://paypal-success",
```
ceriksson, line 72: maybe we should refactor some of this into it's own method since code is repeated? Just thinking loud

ceriksson (approved):
nice! Left a comment, just one thing before merging: how has this been tested?

PR #103, opened 2021-07-12
[CHE-911] Fix payment method config bug

ceriksson:
CHE-911

# What this PR does

https://primerapi.atlassian.net/browse/CHE-911

Fixes a bug that causes payment methods to not load after a payment method type unknown to the SDK has been added to the config.

The fix address the above scenario and filters all payment methods that have not been explicitly enabled for the SDK.

Test this by trying to load the different payment flows in the test app.

# Before merging

_QA_

_Documentation_

# After merging

### DIFF
--- a/Example/PrimerSDK.xcodeproj/xcshareddata/xcschemes/PrimerSDK-Example.xcscheme
+++ b/Example/PrimerSDK.xcodeproj/xcshareddata/xcschemes/PrimerSDK-Example.xcscheme
@@ -96,7 +96,7 @@
       <CommandLineArguments>
          <CommandLineArgument
             argument = "-PrimerDebugEnabled"
-            isEnabled = "NO">
+            isEnabled = "YES">
          </CommandLineArgument>
       </CommandLineArguments>
    </LaunchAction>
--- a/Example/PrimerSDK/CreateClientToken.swift
+++ b/Example/PrimerSDK/CreateClientToken.swift
@@ -12,4 +12,5 @@ import Foundation
 struct CreateClientTokenRequest: Codable {
     let customerId: String
     let customerCountryCode: String?
+    var staging: Bool?
 }
--- a/Example/PrimerSDK/MerchantCheckoutViewController.swift
+++ b/Example/PrimerSDK/MerchantCheckoutViewController.swift
@@ -150,7 +150,7 @@ extension MerchantCheckoutViewController: PrimerDelegate {
         request.httpMethod = "POST"
         request.addValue("application/json", forHTTPHeaderField: "Content-Type")
         
-        let body = CreateClientTokenRequest(customerId: "customer123", customerCountryCode: nil)
+        let body = CreateClientTokenRequest(customerId: "customer123", customerCountryCode: nil, staging: true)
         
         do {
             request.httpBody = try JSONEncoder().encode(body)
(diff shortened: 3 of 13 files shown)


Example/PrimerSDK.xcodeproj/xcshareddata/xcschemes/PrimerSDK-Example.xcscheme
```diff
@@ -96,7 +96,7 @@
       <CommandLineArguments>
          <CommandLineArgument
             argument = "-PrimerDebugEnabled"
-            isEnabled = "NO">
+            isEnabled = "YES">
```
EvansPie, line 99: Let's not commit that. Let's keep as default not logging.

Example/PrimerSDK/CreateClientToken.swift
```diff
@@ -12,4 +12,5 @@ import Foundation
 struct CreateClientTokenRequest: Codable {
     let customerId: String
     let customerCountryCode: String?
+    var staging: Bool?
```
EvansPie, line 15: Is this really needed?

Example/PrimerSDK/MerchantCheckoutViewController.swift
```diff
@@ -150,7 +150,7 @@ extension MerchantCheckoutViewController: PrimerDelegate {
         request.httpMethod = "POST"
         request.addValue("application/json", forHTTPHeaderField: "Content-Type")
         
-        let body = CreateClientTokenRequest(customerId: "customer123", customerCountryCode: nil)
+        let body = CreateClientTokenRequest(customerId: "customer123", customerCountryCode: nil, staging: true)
```
EvansPie, line 153: Same as above. I don't see it used anywhere.

Sources/PrimerSDK/Classes/Data Models/PaymentMethodConfig.swift
```diff
@@ -16,6 +16,18 @@ enum ConfigPaymentMethodType: String, Codable {
     case googlePay = "GOOGLE_PAY"
     case goCardlessMandate = "GOCARDLESS"
     case klarna = "KLARNA"
+    case payNlIdeal = "PAY_NL_IDEAL"
+    
+    case unknown
+    
+    var isEnabled: Bool {
```
EvansPie, line 23: Let's leave that as a quick fix, but create a ticket to abstract it. Shouldn't this be set with config API call?

EvansPie (changes requested):
I added some comments for some minor changes.

Example/PrimerSDK/CreateClientToken.swift
ceriksson: I think we should keep this, in case we want to target staging in the future

Example/PrimerSDK/MerchantCheckoutViewController.swift
ceriksson: It's just for the example app's merchant integration to know we're targeting staging.

Sources/PrimerSDK/Classes/Data Models/PaymentMethodConfig.swift
ceriksson: Not sure I follow.. Could you clarify what you want to do?

Example/PrimerSDK/CreateClientToken.swift
EvansPie: Ok, now I get it. This is just going to add `?staging=true` on the URL, right?

Sources/PrimerSDK/Classes/Data Models/PaymentMethodConfig.swift
EvansPie: What I mean is, shouldn't all other payment methods configured on the dashboard be available? Except for GooglePay of course. Talking about the `isEnabled` bool. @ceriksson

EvansPie (approved):

PR #115, opened 2021-07-26
Feature/dex 207 cardnumber validation

EvansPie:
DEX-207

# What this PR does

Improve card network validation & styling

Styling is only applied on the`PrimerCardNumberTextFieldView` which is going to replace the field in the `FormView`

Run `CardComponentManagerTests` tests, add more card numbers.
Unhide `PrimerCardNumberTextFieldView` in `AppViewController` and test by typing

# Before merging

_QA_

_Documentation_

# After merging

### DIFF
--- a/Sources/PrimerSDK/Classes/Data Models/CardNetwork.swift
+++ b/Sources/PrimerSDK/Classes/Data Models/CardNetwork.swift
@@ -1,0 +1,266 @@
+//
+//  CardNetwork.swift
+//  PrimerSDK
+//
+//  Created by Evangelos Pittas on 29/6/21.
+//
+
+import Foundation
+
+struct CardNetworkValidation {
+    var niceType: String
+    var patterns: [[Int]]
+    var gaps: [Int]
+    var lengths: [Int]
+    var code: CardNetworkCode
+}
+
+struct CardNetworkCode {
+    var name: String
+    var length: Int
+}
+
+public enum CardNetwork: String, CaseIterable {
+    
+    case amex
+    case bancontact
+    case diners
+    case discover
+    case elo
+    case hiper
+    case hipercard
+    case jcb
+    case maestro
+    case masterCard
+    case mir
+    case visa
+    case unionpay
+    case unknown
+    
+    var validation: CardNetworkValidation? {
+        switch self {
+        case .amex:
+            return CardNetworkValidation(
+                niceType: "American Express",
+                patterns: [[34], [37]],
+                gaps: [4, 10],
+                lengths: [15],
+                code: CardNetworkCode(
+                    name: "CID",
+                    length: 4))
+            
+        case .bancontact:
+            return nil
+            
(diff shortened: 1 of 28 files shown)


Sources/PrimerSDK/Classes/Data Models/CardNetwork.swift
```diff
@@ -7,52 +7,265 @@
...
+            } else if pattern.count == 2 {
+                let min = pattern.first!
+                let max = pattern[1]
+                
+                for num in min...max {
+                    let numStr = String(num)
+                    if cardnumber.withoutNonNumericCharacters.hasPrefix(numStr) {
+                        return true
+                    }
+                }
```
xevious78, line 251: [Suggestion]
Feels to me that you can simplify all of these by: 
```
if pattern.count == 1 || pattern.count == 2 {
  let min = pattern.first!
  let max = pattern.count == 2 ? pattern[1] : min

   for num in min...max {
                    let numStr = String(num)
                    if cardnumber.withoutNonNumericCharacters.hasPrefix(numStr) {
                        return true
                    }
     }
}
```

xevious78: However, there's something dangerous with these: the 0 padding.
If the pattern if `[002, 010]`, this will essentially check `2` to `10`, which is not the same!
I've not seen any card starting with `0` though so we should be fine :)

Sources/PrimerSDK/Classes/Extensions & Utilities/StringExtension.swift
```diff
@@ -50,7 +50,24 @@ internal extension String {
     }
 
     var isValidCardNumber: Bool {
-        return count >= 13 && count <= 19 && isValidLuhn
+        let clearedCardNumber = self.withoutNonNumericCharacters
+        
+        if !(clearedCardNumber.count >= 13 && clearedCardNumber.count <= 19 && clearedCardNumber.isValidLuhn) {
+            return false
+        }
```
xevious78, line 57: [Feedback]
I believe you don't need to check the length anymore, it's already checked below.
Similarly, you should do the luhn check after the length is checked: it doesn't make to consider a card non-luhn valid if it's not the right size.

xevious78 (approved):

Sources/PrimerSDK/Classes/Data Models/CardNetwork.swift
EvansPie: Suggestion applied. I thought about the `0`, but then I though there're no cards starting with `0`. I could replace the patterns with strings, but I think it's not needed at the moment.

Sources/PrimerSDK/Classes/Extensions & Utilities/StringExtension.swift
EvansPie: Done

PR #122, opened 2021-07-29
add apaya data models

ceriksson:
Adds Apaya-related data models + exceptions + updates the app state protocol to include Apaya data. This PR will be updated for QA once the other PRs have been merged into this branch.


Sources/PrimerSDK/Classes/User Interface/Checkout/DirectCheckoutViewModel.swift
```diff
@@ -103,6 +103,8 @@ struct PaymentMethodViewModel {
             return ""
         case .klarna:
             return ""
+        case .apaya:
+            return "Pay by mobile"
```
ceriksson, line 107: will turn this into localised string once we agree on the copy.

Sources/PrimerSDK/Classes/User Interface/Primer/PaymentMethodComponent.swift
```diff
@@ -65,6 +65,16 @@ internal class PaymentMethodComponent: PrimerView {
...
             anchorIconView(inRelationToLabel: false)
+        case .apaya:
+            layer.borderWidth = 1
+            layer.borderColor = theme.colorTheme.disabled1.cgColor
+            backgroundColor = theme.colorTheme.main1
+            label.textColor = theme.colorTheme.text1
+            addSubview(label)
+            addSubview(iconView)
+            configureLabel(with: method.toString())
+            anchorLabel()
```
EvansPie, line 76: `anchorLabel` and `anchorIconView` can be unified since they're both `UIView`s and anchoring is for views. The `anchorLabel` naming is not very clear to me, since anchoring is done on another view. On which view is the label anchored?

Sources/PrimerSDK/Classes/Core/Primer/AppState.swift
```diff
@@ -26,10 +26,12 @@ internal protocol AppStateProtocol: class {
     var authorizationToken: String? { get set }
     var customerToken: String? { get set }
     var sessionId: String? { get set }
+    
+    func setApayaResult(_ result: Result<ApayaWebViewResult, ApayaException>)
```
EvansPie, line 30: Are those for testing reasons? Or will the app state have those functions?

Sources/PrimerSDK/Classes/Core/Primer/AppState.swift
```diff
@@ -48,7 +50,17 @@ internal class AppState: AppStateProtocol {
     var authorizationToken: String?
     var customerToken: String?
     var sessionId: String?
+    private var apayaResult: Result<ApayaWebViewResult, ApayaException> = .failure(ApayaException.webViewFlowNotComplete)
```
EvansPie, line 53: This could be optional and be initialized with `nil` value. I know that you init it with `webViewFlowNotComplete` value, but in the case of not having Apaya enabled this doesn't make a lot of sense. I think that `nil` is more suitable since it means that it wasn't initialized.

Sources/PrimerSDK/Classes/Data Models/Apaya.swift
```diff
@@ -0,0 +1,84 @@
+//
+//  Apaya.swift
+//  PrimerSDK
+//
+//  Created by Carl Eriksson on 27/07/2021.
+//
+
+public struct ApayaCreateSessionAPIRequest: Encodable {
```
EvansPie, line 8: This could be refactored in sth like the following. I think it's more readable.

```
struct Apaya {
    struct CreateSessionAPIRequest {
        // ...
    }

    struct CreateSessionAPIResponse {
        // ...
    }
}
```

and be used as:

```
Apaya.CreateSessionAPIRequest()
```

Sources/PrimerSDK/Classes/Data Models/PaymentMethodTokenizationRequest.swift
```diff
@@ -17,6 +17,7 @@ struct PaymentMethodTokenizationRequest: Encodable {
 
 }
 
+// feels like we could polymorph this with a protocol, or at least restrict construcions with a specific factory method for each payment instrument.
```
EvansPie, line 20: Yeah. Wanna try sth like the code below?

```
protocol PaymentInstrumentProtocol {
    var name: String { get }
    // ...
}
protocol OAuthPaymentInstrumentProtocol: PaymentInstrumentProtocol {
    var host: String { get }
    var token: String { get }
}

struct CardPaymentInstrument: PaymentInstrumentProtocol {
    var name: String = "Card"
}

struct ApayaPaymentInstrument: OAuthPaymentInstrumentProtocol {
    var name: String = "Apaya"
    var token: String = "access_token"
    var host: String = "https://..."
}
```

EvansPie (commented):
good job 💪

Sources/PrimerSDK/Classes/Core/Primer/AppState.swift
ceriksson: Could definitely change this param, but wouldn't this mean we have to add a lot of `guard` statements rather than rely purely on switch case?

Sources/PrimerSDK/Classes/Data Models/Apaya.swift
ceriksson: nice!

Sources/PrimerSDK/Classes/Data Models/PaymentMethodTokenizationRequest.swift
ceriksson: I'm hesitant to do this because it would entail significant QA overhead. Think it would be better to create a separate PR for this so that the testing is more focused lest we introduce some bug.

Sources/PrimerSDK/Classes/Core/Primer/AppState.swift
ceriksson: I'm more and more leaning towards this approach, and in the long run we would split up the app state in more manageable stores. The reason is that it simply becomes more testable + the internal workings of the app state becomes the preserve of that class alone. The rest of the SDK shouldn't "see" everything inside this class.

Sources/PrimerSDK/Classes/User Interface/Primer/PaymentMethodComponent.swift
ceriksson: Will look into this and give an update. Generally, this component's logic could be greatly improved. Just not sure what's the best approach. What do you think?

Sources/PrimerSDK/Classes/Data Models/Apaya.swift
```diff
@@ -0,0 +1,84 @@
+//
```
ceriksson, line 1: Important!

Sources/PrimerSDK/Classes/Error Handler/PrimerError.swift
```diff
@@ -13,6 +13,55 @@ internal protocol PrimerErrorProtocol: CustomNSError, LocalizedError {
     var shouldBePresented: Bool { get }
```
ceriksson, line 13: Important!

PR #123, opened 2021-08-02
add apaya service

ceriksson:
Adds payment method service for the apaya flow. Includes one call to create a payment session that is then used to display the web view. This PR will be merged into the main Apaya PR. QA will be done on that branch when everything else has been merged in and the flow is ready.


Sources/PrimerSDK/Classes/Core/Payment Services/ApayaService.swift
```diff
@@ -0,0 +1,41 @@
+//
+//  ApayaService.swift
+//  PrimerSDK
+//
+//  Created by Carl Eriksson on 26/07/2021.
+//
+
+#if canImport(UIKit)
+
+internal protocol ApayaServiceProtocol {
```
EvansPie, line 10: I've noticed that all oauth payment methods (at least until now) have a way to create a payment session. I think that the abstraction should go at this level. Creating a protocol for each service, doesn't abstract anything.

Implementation could be sth like:
```
protocol OAuthServiceProtocol {
    func createPaymentSession(_ completion: @escaping (Result<String, Error>) -> Void)
}

// Then if any of the oauth services have any specifics

protocol ApayaServiceProtocol: OAuthServiceProtocol {
    func doSth()
}

internal class ApayaService: ApayaServiceProtocol {

    func createPaymentSession(_ completion: @escaping (Result<String, Error>) -> Void) {

    }

    func doSth() {

    }
}

// Same goes for the PrimerAPIClientProtocol function apayaCreateSession
// It could be createOAuthSession and get the url as a param.

```

Sources/PrimerSDK/Classes/Core/Payment Services/ApayaService.swift
```diff
@@ -0,0 +1,41 @@
...
+        let state: AppStateProtocol = DependencyContainer.resolve()
+        guard let clientToken = state.decodedClientToken else {
+            return completion(.failure(ApayaException.noToken))
+        }
+        let body = ApayaCreateSessionAPIRequest(locale: "en-GB", itemDescription: "nothing")
+        let api: PrimerAPIClientProtocol = DependencyContainer.resolve()
+        api.apayaCreateSession(clientToken: clientToken, request: body) { result in
+            switch result {
+            case .failure:
+                completion(.failure(KlarnaException.failedApiCall))
```
EvansPie, line 28: Replace KlarnaException with PrimerError

Sources/PrimerSDK/Classes/Core/Payment Services/ApayaService.swift
```diff
@@ -0,0 +1,41 @@
...
+            case .success(let response):
+                log(
+                    logLevel: .info,
+                    message: "\(response)",
+                    className: "ApayaService",
```
EvansPie, line 33: ApayaService can be a var here, in case the class name changes

Sources/PrimerSDK/Classes/Core/Payment Services/ApayaService.swift
```diff
@@ -0,0 +1,41 @@
...
+                    function: "createPaymentSession"
```
EvansPie, line 34: Same as above, i.e. `#function`

Sources/PrimerSDK/Classes/Services/API/Primer/PrimerAPI.swift
```diff
@@ -39,6 +39,8 @@ internal extension PrimerAPI {
              .klarnaFinalizePaymentSession(let clientToken, _):
             guard let urlStr = clientToken.coreUrl else { return nil }
             return urlStr
+        case .apayaCreateSession:
```
EvansPie, line 42: Remember to remove when there's a backend

EvansPie (approved):
I've approved, but see my comments.

Sources/PrimerSDK/Classes/Core/Payment Services/ApayaService.swift
ceriksson: great idea! will try it out

Sources/PrimerSDK/Classes/Core/Payment Services/ApayaService.swift
```diff
@@ -0,0 +1,45 @@
+//
```
ceriksson, line 1: Important!

PR #126, opened 2021-08-03
add apaya flow

ceriksson:
DEX-191

# What this PR does

This adds the UI/flow logic for the Apaya vaulting flow. If Apaya has been added as a payment method, the user should be presented with the option to "Pay by mobile" by using their phone bill as a vaulted payment method.

This is a web view based flow and relies heavily on the redirect URL received from the Primer platform API.

https://user-images.githubusercontent.com/42890953/128699604-9f8db733-629a-4bf3-81e5-e0546f34de47.mp4

List & explain ...

Run in appetize, firebase, or download and run example app locally. The main thing to test is that each scenario completes properly and that there are no crashes or frozen states. It's also important to make sure that all results (error, success, cancelled) are always communicated back to the developer & that this communication makes sense.

# Before merging

_QA_

_Documentation_

# After merging


Sources/PrimerSDK/Classes/User Interface/OAuth/OAuthViewController.swift
```diff
@@ -23,42 +23,39 @@ internal class OAuthViewController: PrimerViewController {
     deinit {
         log(logLevel: .verbose, message: "🧨 destroyed: \(self.self)")
     }
```
ceriksson, line 25: I will revert this file to its original state. Believe it makes more sense to maintain smaller separate view controllers with better abstraction. Wdyt?

Sources/PrimerSDK/Classes/User Interface/Checkout/VaultCheckoutViewController.swift
```diff
@@ -91,7 +91,7 @@ extension VaultCheckoutViewController: UITableViewDelegate, UITableViewDataSourc
             cell.accessibilityIdentifier = "payment_method_table_view_paypal_cell"
         case .goCardlessMandate:
             cell.accessibilityIdentifier = "payment_method_table_view_direct_debit_cell"
-        case .paymentCard:
+        case .paymentCard, .apaya:
```
EvansPie, line 94: It should have its own identifier for UI tests, i.e. `payment_method_table_view_apaya`

Sources/PrimerSDK/Classes/User Interface/OAuth/OAuthViewController.swift
EvansPie: In the future I'd like to completely remove this controller, which is actually presented between the previous view controller and the webview controller.

Let it be like that for now.

Sources/PrimerSDK/Classes/User Interface/OAuth/OAuthViewController.swift
```diff
@@ -23,42 +23,39 @@ internal class OAuthViewController: PrimerViewController {
...
+    
+    func renderSpinner() {
+        view.addSubview(indicator)
+        let theme: PrimerThemeProtocol = DependencyContainer.resolve()
+        indicator.color = theme.colorTheme.disabled1
+        indicator.translatesAutoresizingMaskIntoConstraints = false
+        indicator.centerYAnchor.constraint(equalTo: view.centerYAnchor).isActive = true
```
EvansPie, line 32: Effectively, you're adding those constraints each time you call `renderSpinner()`. You won't have constraints conflicts though. Reasons are:

- It's the same constraints
- `renderSpinner()` is only called once from the `viewDidLoad`

Sources/PrimerSDK/Classes/User Interface/OAuth/OAuthViewController.swift
```diff
@@ -87,52 +86,59 @@ internal class OAuthViewController: PrimerViewController {
...
-                _ = ErrorHandler.shared.handle(error: err)
-                router.show(.error(error: err))
-                
-            } else {
-                guard let host = self?.host else {
-                    let error = PrimerError.failedToLoadSession
-                    router.show(.error(error: error))
-                    return
+        if (host == .apaya) {
+            let webViewController = ApayaWebViewController()
```
EvansPie, line 90: I'm not sure why we'd need another class for Apaya's webview controller. `klarnaWebviewCompletion` can be optional, and also renamed to `webviewCompletion`.

Other than that, their functionalities are the same I believe. You can just init the controller with the view model needed, then the controller will grab the `url` (or any other information needed) from the view model. 

The `completionHandler` can also be part of the view model, and be removed from the webview controller.

EvansPie (approved):
lgtm, I've left some comments (maybe future work tho)

Sources/PrimerSDK/Classes/User Interface/Checkout/VaultCheckoutViewController.swift
ceriksson: 👍  will add

Sources/PrimerSDK/Classes/User Interface/OAuth/OAuthViewController.swift
ceriksson: Yeah! I think once we compare them we'll be able to figure out what's the best way. I want to avoid situations were these view controllers become too bloated, we should raise this in a separate ticket probably.

Sources/PrimerSDK/Classes/User Interface/OAuth/OAuthViewController.swift
ceriksson: 👍

Sources/PrimerSDK/Classes/User Interface/OAuth/PrimerLoadWebViewController.swift
```diff
@@ -0,0 +1,98 @@
+//
```
ceriksson, line 1: important!

Sources/PrimerSDK/Classes/User Interface/OAuth/PrimerLoadWebViewModel.swift
```diff
@@ -0,0 +1,136 @@
+//
```
ceriksson, line 1: important!

Sources/PrimerSDK/Classes/User Interface/OAuth/PrimerWebViewController.swift
```diff
@@ -0,0 +1,80 @@
+//
```
ceriksson, line 1: important!

Sources/PrimerSDK/Classes/User Interface/OAuth/PrimerWebViewModel.swift
```diff
@@ -0,0 +1,32 @@
+//
```
ceriksson, line 1: important!

PR #127, opened 2021-08-03
Feature/dex 205 use autolayout

EvansPie:
DEX-205 Use auto-layout and child containers

# What this PR does

Refactors the UI's presentation logic.

Implementation is not finished yet.

- Present a primer window
- Window's root view controller contains a child view controller, which serves as the container for our content.
- The container view controller contains a navigation controller which will hold our navigation stack.
- Each of navigation controller's children is of type PrimerContainerViewController, which integrates (and enables) a scrollview if the content is larger than the screen.

Use one of last two buttons to present Primer's view (checkout or vault)

# Before merging

_QA_

_Documentation_

# After merging


ceriksson (commented):
Awesome, just gotta align on the UI. I'd ask a 2nd person review as well, then we should be good. Left some comments, nothing major.

PR #139, opened 2021-09-09
Feature/dex 427 rename main functions

EvansPie:
DEX-427

# What this PR does

Deprecates `showCheckout`, adds & exposes `showUniversalCheckout`, `showVaultManager` and `showPaymentMethod` functions.

Screenshot 2021-09-09 at 4 32 01 PM

# Before merging

_QA_

_Documentation_

# After merging


Sources/PrimerSDK/Classes/Core/Primer/Primer.swift
```diff
@@ -175,6 +176,76 @@ public class Primer {
...
+            guard let self = self else { return }
+            self.root = RootViewController()
+            guard let root = self.root else { return }
+            let router: RouterDelegate = DependencyContainer.resolve()
+            router.setRoot(root)
+            viewController.present(root, animated: true)
+        }
+    }
+    
+    public func showPaymentMethod(_ paymentMethod: ConfigPaymentMethodType, on viewController: UIViewController, with clientToken: String? = nil) {
```
xevious78, line 208: [Feedback]
We need a flag for vault vs. non vault :)
"showPaymentMethod" can be for vaulting or for checkout 👍

Sources/PrimerSDK/Classes/Error Handler/ErrorHandler.swift
```diff
@@ -19,87 +19,19 @@ internal class ErrorHandler {
 
         if let networkServiceError = error as? NetworkServiceError {
             switch networkServiceError {
```
ceriksson, line 21: what's happening here?

Sources/PrimerSDK/Classes/Core/Primer/Primer.swift
```diff
@@ -175,6 +176,85 @@ public class Primer {
...
+            guard let self = self else { return }
+            self.root = RootViewController()
+            guard let root = self.root else { return }
+            let router: RouterDelegate = DependencyContainer.resolve()
+            router.setRoot(root)
+            viewController.present(root, animated: true)
+        }
+    }
+    
+    public func showPaymentMethod(_ paymentMethod: ConfigPaymentMethodType, withIntent intent: PrimerSessionIntent, on viewController: UIViewController, with clientToken: String? = nil) {
```
ceriksson, line 208: do we need unit tests for this method?

ceriksson (commented):
👍 just too questions, otherwise good

Sources/PrimerSDK/Classes/Core/Primer/Primer.swift
```diff
@@ -175,6 +176,85 @@ public class Primer {
...
+            guard let self = self else { return }
+            self.root = RootViewController()
+            guard let root = self.root else { return }
+            let router: RouterDelegate = DependencyContainer.resolve()
+            router.setRoot(root)
+            viewController.present(root, animated: true)
+        }
+    }
+    
+    public func showVaultManager(on viewController: UIViewController, clientToken: String? = nil) {
```
xevious78, line 194: Doesn't seem right that `clientToken` is not used in the function 🤔 
I think we should just remove it for now.

Sources/PrimerSDK/Classes/Data Models/PrimerInternalSessionFlow.swift
```diff
@@ -128,3 +128,8 @@ internal enum PrimerInternalSessionFlow {
         }
     }
 }
+
+public enum PrimerSessionIntent: String {
```
xevious78, line 132: [Feedback]
Feels weird that a `public` enum is in Primer*Internal*SessionFlow.

Sources/PrimerSDK/Classes/Error Handler/PrimerError.swift
```diff
@@ -485,6 +488,13 @@ enum PrimerError: PrimerErrorProtocol {
                                      bundle: Bundle.primerResources,
                                      value: "Amount cannot be null for order items with isPending == false",
                                      comment: "Amount cannot be null for order items with isPending == false - Primer error message")
+            
+        case .intentNotSupported(let intent, let paymentMethodType):
+            return NSLocalizedString("primer-error-message-intent-not-supported",
+                                     tableName: nil,
+                                     bundle: Bundle.primerResources,
+                                     value: "Intent is not supported",
+                                     comment: "Intent is not supported - Primer error message") + "\(paymentMethodType.rawValue).\(intent.rawValue)"
```
xevious78, line 497: I know it's an edge case but do you think it makes sense to call it just `Intent`, I believe it's better to call `PrimerSessionIntent not supported`?

xevious78 (commented):
Some comments :)
I think we should have some unit tests for the main functions - we need to be extra sure that we don't break them 🤔

Sources/PrimerSDK/Classes/Core/Primer/Primer.swift
EvansPie: `intent` (checkout or vault) has been added

Sources/PrimerSDK/Classes/Core/Primer/Primer.swift
EvansPie: Fixed. Bear in mind that this can be used in two ways:

1. `.showVaultManager(on: vc)`
2. `.showVaultManager(on: vc, clientToken: "a_token")`

The second one will avoid a new request for token.

Sources/PrimerSDK/Classes/Data Models/PrimerInternalSessionFlow.swift
EvansPie: Yep, naming wasn't the best. Renamed the file.

Sources/PrimerSDK/Classes/Core/Primer/Primer.swift
EvansPie: Done

Sources/PrimerSDK/Classes/Error Handler/PrimerError.swift
EvansPie: Done

PR #143, opened 2021-09-14
Fix Amount String

ceriksson:
CHE-

The amount string in the main checkout view displayed incorrectly for some currencies.

Fixed this + added some unit tests

Describe ... (Features / UI / logic changes)

List & explain ...

# Before merging

_QA_

_Documentation_

# After merging

### DIFF
--- a/Sources/PrimerSDK/Classes/Extensions & Utilities/IntExtension.swift
+++ b/Sources/PrimerSDK/Classes/Extensions & Utilities/IntExtension.swift
@@ -7,11 +7,31 @@
 
 internal extension Int {
     func toCurrencyString(currency: Currency) -> String {
-        var double = Double(self)
-        if (!currency.isZeroDecimal) {
-            double /= 100
+        let nf = NumberFormatter()
+        nf.usesGroupingSeparator = true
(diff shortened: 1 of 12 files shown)


Sources/PrimerSDK/Classes/Extensions & Utilities/IntExtension.swift
```diff
@@ -0,0 +1,24 @@
...
+//  IntExtension.swift
+//  PrimerSDK
+//
+//  Created by Carl Eriksson on 14/09/2021.
+//
+
+internal extension Int {
+    func toCurrencyString(currency: Currency) -> String {
+        switch currency {
+        case .USD: return String(format: "$%.2f", Float(self) / 100)
```
EvansPie, line 11: This (i.e. `String(format: "$%.2f", Float(self) / 100)`) can lead to wrong amount, due to floating point.

Can you check how to do that using a `NumberFormatter`?

```
let formatter = NumberFormatter()
formatter.numberStyle = .currency
```

Also an idea, we could add the currency symbols in the `Currency` structure we already have, and use them from there.

EvansPie (changes requested):
@ceriksson check my comment

PR #147, opened 2021-09-21
Feature/dex 482 apaya test app

EvansPie:
DEX-482 Apaya Test App

# What this PR does

Create test app
Refactors some of the Apaya logic

List & explain ...

# Before merging

_QA_

_Documentation_

# After merging


Sources/PrimerSDK/Classes/Data Models/Apaya.swift
```diff
@@ -82,4 +82,91 @@ public struct Apaya {
             )
         }
     }
+    
+    class ViewModel {
```
semirp, line 86: Are we sure we want this to be in the SDK too? I believe either API will return us the `carrierName` or the DEV has to do this.

Sources/PrimerSDK/Classes/User Interface/OAuth/PrimerLoadWebViewModel.swift
```diff
@@ -154,6 +154,7 @@ internal class ApayaLoadWebViewModel: PrimerLoadWebViewModelProtocol {
         let clientTokenService: ClientTokenServiceProtocol = DependencyContainer.resolve()
         clientTokenService.loadCheckoutConfig { [weak self] (error) in
             if let error = error {
+                Primer.shared.delegate?.checkoutFailed?(with: error)
```
ceriksson, line 157: are we sure we only call this method once?

Sources/PrimerSDK/Classes/Data Models/Apaya.swift
```diff
@@ -82,4 +82,91 @@ public struct Apaya {
...
+                  let carrier = Apaya.Carrier(mcc: mcc, mnc: mnc)
+            else { return nil }
+            
+            self.carrier = carrier
+            self.hashedIdentifier = paymentMethod.paymentInstrumentData?.hashedIdentifier
+        }
+        
+    }
+
+    enum Carrier: String, Codable {
```
ceriksson, line 104: do we need unit tests for these?

ceriksson (commented):
nice! left a question or two. Gonna QA the checkout, after that we should be good.

Sources/PrimerSDK/Classes/Data Models/Apaya.swift
EvansPie: Originally this was supposed to be only in the Example app's side. However, I added it on the SDK as well cause we might need it. I can add some tests.

Sources/PrimerSDK/Classes/User Interface/OAuth/PrimerLoadWebViewModel.swift
EvansPie: I believe so, it is not called from `navigate`, nor the `router`.

PR #223, opened 2021-11-22
Feature/dex 859 release config v3

EvansPie:
DEX-859

# What this PR does

Consume config v3 and deprecate properties in `PrimerSettings`.

List & explain ...

# Before merging

_QA_

_Documentation_

# After merging

### DIFF
--- a/Sources/PrimerSDK/Classes/Data Models/OrderItem.swift
+++ b/Sources/PrimerSDK/Classes/Data Models/OrderItem.swift
@@ -10,6 +10,7 @@
 import Foundation
 import PassKit
 
+@available(*, deprecated, message: "Set the order items in the client session")
 public struct OrderItem: Codable {
     public let name: String
     public let unitAmount: Int?
--- a/Sources/PrimerSDK/Classes/Data Models/OrderItem.swift
+++ b/Sources/PrimerSDK/Classes/Data Models/OrderItem.swift
@@ -10,8 +10,9 @@
 import Foundation
 import PassKit
 
-@available(*, deprecated, message: "Set the order items in the client session")
+@available(*, deprecated, message: "Set the order items in the client session with POST /client-session. See documentation here: https://primer.io/docs/api#tag/Client-Session")
 public struct OrderItem: Codable {
+    
     public let name: String
     public let unitAmount: Int?
     public let quantity: Int
--- a/Sources/PrimerSDK/Classes/Data Models/PrimerSettings.swift
+++ b/Sources/PrimerSDK/Classes/Data Models/PrimerSettings.swift
@@ -6,12 +6,16 @@ public typealias TokenizationSuccessCallBack = (_ paymentMethodToken: PaymentMet
 public typealias CheckoutDismissalCallback = () -> Void
 
 internal protocol PrimerSettingsProtocol {
+    @available(*, deprecated, message: "Use client session")
     var amount: Int? { get }
+    @available(*, deprecated, message: "Use client session")
     var currency: Currency? { get }
     var merchantIdentifier: String? { get }
+    @available(*, deprecated, message: "Use client session")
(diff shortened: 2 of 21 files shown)


Sources/PrimerSDK/Classes/Data Models/OrderItem.swift
```diff
@@ -10,6 +10,7 @@
 import Foundation
 import PassKit
 
+@available(*, deprecated, message: "Set the order items in the client session")
```
xevious78, line 13: Not sure it's clear that the client session is something on the backend. Developers would first look at a SDK settings called "client session".

Maybe something like this?
`Set the order items in the client session with POST /client-session`

Sources/PrimerSDK/Classes/Data Models/PrimerSettings.swift
```diff
@@ -6,12 +6,16 @@ public typealias TokenizationSuccessCallBack = (_ paymentMethodToken: PaymentMet
 public typealias CheckoutDismissalCallback = () -> Void
 
 internal protocol PrimerSettingsProtocol {
+    @available(*, deprecated, message: "Use client session")
```
xevious78, line 9: Again "Use client session" doesn't help much. Could you rephrase?
"Pass this in the client session with POST /client-session"

xevious78 (approved):
could you improve the deprecation messages? :)

Sources/PrimerSDK/Classes/Data Models/OrderItem.swift
ceriksson: We could also add a url link to the relevant part in the API documentation.

Sources/PrimerSDK/Classes/Data Models/Throwable.swift
```diff
@@ -0,0 +1,35 @@
...
+//  PrimerSDK
+//
+//  Created by Evangelos on 22/11/21.
+//
+
+import Foundation
+
+#if canImport(UIKit)
+
+enum Throwable<T: Decodable>: Decodable {
```
ceriksson, line 12: nice!

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/KlarnaTokenizationViewModel.swift
```diff
@@ -147,13 +147,13 @@ class KlarnaTokenizationViewModel: PaymentMethodTokenizationViewModel, ExternalP
...
             
-            if settings.orderItems.isEmpty {
+            if (settings.orderItems ?? []).isEmpty {
                 let err = KlarnaException.missingOrderItems
                 _ = ErrorHandler.shared.handle(error: err)
                 throw err
             }
             
-            if !settings.orderItems.filter({ $0.unitAmount == nil }).isEmpty {
+            if !(settings.orderItems ?? []).filter({ $0.unitAmount == nil }).isEmpty {
```
ceriksson, line 156: maybe the orderItems should default to an empty array so that we don't need to null check each time?

ceriksson (commented):
looks great, just the comments to address :)

Sources/PrimerSDK/Classes/Data Models/OrderItem.swift
EvansPie: Makes sense. I'll include the URL as well.

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/KlarnaTokenizationViewModel.swift
EvansPie: I did that on purpose. This way we know that they haven't been set as long as they're nil, or that they have been set to an empty array (which means we have a response).

PR #256, opened 2021-12-17
Feature/dex 999 analytics

EvansPie:
DEX-999

# What this PR does

- Adds analytics on the SDK

List & explain ...

# Before merging

_QA_

_Documentation_

# After merging


Sources/PrimerSDK/Classes/Core/Analytics/AnalyticsService.swift
```diff
@@ -0,0 +1,220 @@
...
+//  Created by Evangelos on 13/12/21.
+//
+
+#if canImport(UIKit)
+
+import Foundation
+
+extension Analytics {
+    
+    internal class Service {
```
ceriksson, line 14: could we make this protocol-based? There are some places in the code (utils, extensions, etc.) where we reference this service, but it feels like it would be better to inject it as a protocol so that it's easier to write unit tests, etc.

Sources/PrimerSDK/Classes/Extensions & Utilities/StringExtension.swift
```diff
@@ -130,7 +152,18 @@ internal extension String {
             return count == numberOfDigits
         }
         
-        return count > 2 && count < 5
+        let isValid = count > 2 && count < 5
+        
+        if !isValid {
+            let event = Analytics.Event(
```
ceriksson, line 158: actually, would it maybe be better to call the analytics service inside view models & other services? It feels a bit heavy to put all of this inside utility methods

ceriksson (commented):
really nice! Just got two questions regarding the injection of analytics service.

Sources/PrimerSDK/Classes/Core/Analytics/AnalyticsService.swift
EvansPie: @ceriksson yes, and this one makes sense to be added in the dependency container.

Sources/PrimerSDK/Classes/Extensions & Utilities/StringExtension.swift
EvansPie: You're right, fixed.

PR #263, opened 2021-12-27
Feature/dex 618 card form fields autofocus

EvansPie:
DEX-618

# What this PR does

- Autofocuses on the next card form field if user has arrived on the max length, and also text has passed validation.
- Removes error text when field is empty

List & explain ...

# Before merging

_QA_

_Documentation_

# After merging

### DIFF
--- a/Sources/PrimerSDK/Classes/User Interface/Text Fields/PrimerCVVFieldView.swift
+++ b/Sources/PrimerSDK/Classes/User Interface/Text Fields/PrimerCVVFieldView.swift
@@ -50,7 +50,11 @@ public final class PrimerCVVFieldView: PrimerTextFieldView {
         
         switch validation {
         case .valid:
-            delegate?.primerTextFieldView(self, isValid: true)
+            if let cvvLength = cardNetwork.validation?.code.length, newText.count == cvvLength {
+                delegate?.primerTextFieldView(self, isValid: true)
+            } else {
+                delegate?.primerTextFieldView(self, isValid: nil)
+            }
         case .invalid:
             if let cvvLength = cardNetwork.validation?.code.length, newText.count == cvvLength {
                 delegate?.primerTextFieldView(self, isValid: false)
--- a/Sources/PrimerSDK/Classes/User Interface/Text Fields/PrimerCVVFieldView.swift
+++ b/Sources/PrimerSDK/Classes/User Interface/Text Fields/PrimerCVVFieldView.swift
@@ -45,15 +45,22 @@ public final class PrimerCVVFieldView: PrimerTextFieldView {
             validation = .notAvailable
         }
         
+        primerTextField._text = newText
+        primerTextField.text = newText
+        
         switch validation {
         case .valid:
             delegate?.primerTextFieldView(self, isValid: true)
+        case .invalid:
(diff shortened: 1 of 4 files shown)


Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/FormTokenizationViewModel.swift
```diff
@@ -436,14 +436,41 @@ extension CardFormPaymentMethodTokenizationViewModel: PrimerTextFieldViewDelegat
...
+            // We know for sure that the text is valid, but could change to invalid if the user continues typing.
+            if primerTextFieldView is PrimerCardNumberFieldView {
+                cardNumberContainerView.errorText = nil
+                _ = expiryDateField.becomeFirstResponder()
+            } else if primerTextFieldView is PrimerExpiryDateFieldView {
+                expiryDateContainerView.errorText = nil
+                _ = cvvField.becomeFirstResponder()
+            } else if primerTextFieldView is PrimerCVVFieldView {
+                cvvContainerView.errorText = nil
+                _ = cardholderNameField.becomeFirstResponder()
```
xevious78, line 460: I'd suggest spliting this into two functions: `updateErrorText` and `focusNextFieldIfNeeded`
It'd be easier to read.

That being said,  I don't find it particularly scalable. I think you should have an array telling you the order of the text fields - this will help you define what is the next field to focus.

Sources/PrimerSDK/Classes/User Interface/Text Fields/PrimerCVVFieldView.swift
```diff
@@ -45,15 +45,22 @@ public final class PrimerCVVFieldView: PrimerTextFieldView {
...
+        primerTextField.text = newText
+        
         switch validation {
         case .valid:
             delegate?.primerTextFieldView(self, isValid: true)
+        case .invalid:
+            if let cvvLength = cardNetwork.validation?.code.length, newText.count == cvvLength {
+                delegate?.primerTextFieldView(self, isValid: false)
+            } else {
+                delegate?.primerTextFieldView(self, isValid: nil)
```
xevious78, line 58: [Q]
Just so that I understand `isValid` - is that correct?
* `false` means "it's not valid"
* `nil` means "we don't now yet"
* `true` means "it's valid"

xevious78 (commented):
A few comments.

For some reasons, Amex (which has a CVV length of 4) does the autofocus of CVV after 3 digits.
Could look into that?

Sources/PrimerSDK/Classes/User Interface/Text Fields/PrimerCVVFieldView.swift
EvansPie: yes

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/FormTokenizationViewModel.swift
EvansPie: Fixed the first one. About the second one, I think it's better to be implemented along with the dynamic forms that will have a position.

PR #265, opened 2021-12-28
Feature/dex 1027 track error events

EvansPie:
DEX-1027

# What this PR does

- Refactors error structure according to https://www.notion.so/primerio/Errors-DX-1b5484022bca4aa09b4a14009fcb7857
- Creates appropriate errors
- Create error events that are sent via the analytics service

List & explain ...

# Before merging

_QA_

_Documentation_

# After merging


Sources/PrimerSDK/Classes/Core/3DS/Networking/PrimerAPIClient+3DS.swift
```diff
@@ -13,69 +13,30 @@ extension PrimerAPIClient {
...
+            case .failure(let err):
+                completion(.failure(err))
             }
         }
         
     }
     
 }
 
-extension MockPrimerAPIClient {
```
EvansPie, line 44: This was moved to a separate file

Sources/PrimerSDK/Classes/Core/Crypto/AES256.swift
```diff
@@ -0,0 +1,172 @@
+//
```
EvansPie, line 1: Even though I'm not encrypting analytics event anymore, maybe we'll need to encrypt/decrypt data in the future

Sources/PrimerSDK/Classes/Data Models/ClientToken.swift
```diff
@@ -60,15 +60,21 @@ struct DecodedClientToken: Codable {
     
     func validate() throws {
         if accessToken == nil {
-            throw PrimerError.clientTokenNull
+            let err = PrimerError.invalidClientToken(userInfo: ["file": #file, "class": "\(Self.self)", "function": #function, "line": "\(#line)"])
```
ceriksson, line 63: wouldn't a developer want to know whether the client token expired as opposed to it being missing for some reason?

ceriksson (commented):
looks good, but I would just take an extra look to ensure actionability for any error we pass to the integrating developer. I.e. if something is invalid, I want to know why so that I can fix it.

ceriksson (approved):

PR #273, opened 2022-01-12
[DEX-1044] Implemented the script to fetch the translations from Phrase

emirbostan:
DEX-1044

# What this PR does

This PR introduces a new NodeJS script that fetches all the translations available on Phrase and create the new `Localizable.strings` files for each locale.

List & explain ...

# Before merging

_QA_

_Documentation_

# After merging


tools/phrase/index.js
```diff
@@ -0,0 +1,129 @@
...
+    return;
+  }
+
+  ensureDirectoryExistence(dirname);
+  fs.mkdirSync(dirname);
+};
+
+const formatTranslationString = (translation) => {
+  return translation
+    .replace(/'/g, "\\'")
```
emirbostan, line 43: Is escaping needed?

tools/phrase/index.js
```diff
@@ -0,0 +1,129 @@
...
+    .replace(/"/g, '\\"');
+};
+
+const saveLocalizableStrings = (filePath, data) => {
+  ensureDirectoryExistence(filePath);
+  i18nStringsFiles.writeFileSync(filePath, data, { encoding: 'UTF-16', wantsComments: true });
+};
+
+const getLocales = () =>
+  phraseRequestInstance.get('/locales', { params: { page: 1, per_page: 100 } });
```
EvansPie, line 53: is there any way this wouldn't work in the future?

tools/phrase/env.base
```diff
@@ -0,0 +1,2 @@
+PHRASE_ACCESS_KEY=
```
EvansPie, line 1: Would we have to enter the keys before running it? This could easily lead to someone forgetting to remove them, why don't you gitignore this file? You'd also have to remove it from git now that it's already tracked.

EvansPie (commented):
Left a couple of comments. Also, can you check if you can add an Xcode script phase that runs this script on build, or preferably archive.

tools/phrase/env.base
emirbostan: One has to create a copy of this file as `.env` and put the Phrase Project ID and Access Key in that file.

We don't track `.env` file on git, so I guess we won't have any problems.

PR #290, opened 2022-02-02
Feature/dex 1082 checkout components

EvansPie:
DEX-1082

# What this PR does

Poc of checkout components

List & explain ...

# Before merging

_QA_

_Documentation_

# After merging


Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerCheckoutComponents.swift
```diff
@@ -0,0 +1,222 @@
...
+
+#if canImport(UIKit)
+
+import UIKit
+
+public class PrimerCheckoutComponents {
+    
+    public static var delegate: PrimerCheckoutComponentsDelegate?
+    
+    public static func listAvailablePaymentMethods(forSession clientToken: String, completion: @escaping ([PaymentMethodConfigType]?, Error?) -> Void) {
```
d4r1091, line 16: Not sure about the grammar and convention we utilise here so perhaps is going to be a pointless comment.
-----
Calling this function I expect to receive a list/array of "available payment methods" which I suppose is a list of models. What I get as a result is a list of "payment method **types**".
As per the function below being more explicit into the kind of list, I'm receiving, I was wondering where it would be better to have:

```suggestion
    public static func listAvailablePaymentMethodTypes(forSession clientToken: String, completion: @escaping ([PaymentMethodConfigType]?, Error?) -> Void) {
```

Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerCheckoutComponents.swift
```diff
@@ -0,0 +1,222 @@
...
+        case .payPal:
+            return []
+        case .xfers:
+            return []
+        case .other(let rawValue):
+            return []
+        }
+    }
+    
+    public static func getButton(for paymentMethodType: PaymentMethodConfigType) -> UIButton? {
```
d4r1091, line 103: Same as above as precondition 😅 
----
This looks like a "factory method", so unless defined in our Primer's convention and styles (e.g. "We at primer write API with the same method signature among platform"), I guess here we could use:

```suggestion
    public static func makeButton(paymentMethodType: PaymentMethodConfigType) -> UIButton? {
```

Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerCheckoutComponentsProtocols.swift
```diff
@@ -0,0 +1,171 @@
+//
```
d4r1091, line 1: This is a perfect starting point as a POC and implementation. Like the approach of separating responsibilities.
Was thinking about separating the below "validators" into appropriate classes so that an input form can be "composable" and accept different *Validator, defaulting to the standard ones required for card validation.

e.g.

```
CardNumberFormatValidator
CardNumberExpiryDateValidator
CardNumberCVVValidator
```

and so on...

d4r1091 (commented):
Thanks for letting me have a look @EvansPie 💪 
Great work 👍

d4r1091 (approved):
As we are on the same page for futures' improvements, I'm happy with that 👍

Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerCheckoutComponents.swift
EvansPie: Done!

Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerCheckoutComponents.swift
EvansPie: Done!

d4r1091:
Closing as no longer needed.

PR #298, opened 2022-02-10
Feature/dex 1082 checkout components c

EvansPie:
DEX-1082

# What this PR does

Refactor API

List & explain ...

# Before merging

_QA_

_Documentation_

# After merging


Example/PrimerSDK/CreateClientToken.swift
```diff
@@ -11,6 +11,48 @@ import PrimerSDK
...
+            return 1
+        case .sandbox:
+            return 2
+        case .staging:
+            return 3
+        case .production:
+            return 4
+        }
+    }
+
```
d4r1091, line 46: I was wondering whether the value of _simplifying_ this enum a bit more.
I wouldn't do that if we planned to handle the `default` value with something different.
But since we are sending a `fatalError()` (that should never happen in any case) I thought that we could benefit from swift a bit more.
In this case, whenever we add or remove a `case` we will have less code to deal with.

If you don't see the value in it, I'm fine with the current implementation as this is the sample app.

```suggestion
enum Environment: String, Codable, CaseIterable {
    case local, dev, sandbox, staging, production
    
    init(intValue: Int) {
        self = Environment.allCases[intValue]
    }
    
    var intValue: Int {
        Environment.allCases.firstIndex(of: self) ?? 0
    }    
```

Example/PrimerSDK/MerchantCheckoutViewController.swift
```diff
@@ -86,17 +85,8 @@ class MerchantCheckoutViewController: UIViewController {
         )
 
         
-        Primer.shared.configure(settings: generalSettings)
-        Primer.shared.configure(theme: CheckoutTheme.primer)
+        Primer.shared.configure(settings: generalSettings, theme: CheckoutTheme.primer)
         Primer.shared.delegate = self
```
d4r1091, line 89: I'm quite sure I've added it as part of the [Notion suggestion](https://www.notion.so/Checkout-Components-API-b917503bb5ca4b7da3d2af031e8208f9?d=b9fa7eb6a59243199b0db17a928f7bfc#5b655d7552354f9d92956973fb02eb60) too - was wondering if we'd add the `delegate` as part of the configuration func (with default value to `nil` perhaps)

Example/PrimerSDK/MerchantCheckoutViewController.swift
```diff
@@ -470,17 +274,35 @@ extension MerchantCheckoutViewController: PrimerDelegate {
             )
         )
         
-        requestClientSession(requestBody: clientSessionRequestBody, completion: completion)
+        let networking = Networking()
+        networking.requestClientSession(requestBody: clientSessionRequestBody) { (clientToken, err) in
+            if let err = err {
+                print(err)
+                let merchantErr = NSError(domain: "merchant-domain", code: 1, userInfo: [NSLocalizedDescriptionKey: "Failed to fetch client token"])
```
d4r1091, line 281: Was wondering whether to move all these hardcoded strings into a separate file...
literally polishing the SDK and having a unique source of truth `Strings.swift`
Even though this is the Example - I think it's good to provide consistency in "the way we code" between the SDK and Example app

For later...but worth to keep in mind

d4r1091 (commented):
Apart from reviewing the way we handle Strings (for later of course). I'm happy with the GREAT refactoring you've been carrying over @EvansPie . Just left a little suggestion for handling the `Environment` in the Example app.

d4r1091:
Closing as no longer needed.

PR #305, opened 2022-02-14
DEX-1082 Reviewed PrimerButton and minor fixes

d4r1091:
DEX-1082 -> [DEX-1266](https://primerapi.atlassian.net/browse/DEX-1266)

# What this PR does

- Refactored/Unified `PrimerOldButton` and `PrimerButton` into a single `UIButton` > `PrimerButton`
- Added Safe Area to` Main.storyboard`
- Refactored MerchantCheckoutViewController
    - Reviewed constraints and stackviews to simplify further designs
    - Impact on the UI minimum (better alignment, overall better UI hierarchy structure)
    - Impact on the development, major (quicker UI debugging and fewer chances of runtime constraints crash)
    - Now it handles correctly all range of devices matching our SDK current minimum target (from 4S up)
- Added `Strings.swift` file with an example of merging all strings into a single file.
    - Would love to start keeping this as per the default style when dealing with localization

### Things I decided to postpone:
- The positioning for the images with respect to buttons because of the upcoming [RTL ticket](https://primerapi.atlassian.net/browse/DEX-1288).
- The review of the Themeing structure

Added `@@IBInspectable` and `@IBDesignable` to expose the `PrimerButton` to live Interface Builder editing.

App View Controller Scene

# Before merging

_QA_

_Documentation_

# After merging


d4r1091:
This one can be now reviewed as I addressed your point in regard to making the button fully customizable without having to rely on a Theme entity from PrimerSDK @EvansPie

d4r1091:
Also providing more context based on the JIRA message:


> I’ve gone through a check and it looks like the “increased padding” in the RTL case is due to the nature of the iOS Toolbar insets. As you can notice, for RTL we have the “mirror effect” ad the back button :arrow_backward: is reversed.

Screen Shot 2022-03-02 at 19 13 48

Screen Shot 2022-03-02 at 19 14 53

Sources/PrimerSDK/Classes/Extensions & Utilities/PrimerButton.swift
```diff
@@ -13,59 +13,183 @@ import UIKit
...
-        topAnchor.constraint(equalTo: view.topAnchor, constant: top).isActive = true
-        bottomAnchor.constraint(equalTo: view.bottomAnchor, constant: bottom).isActive = true
-        leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: leading).isActive = true
-        trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: trailing).isActive = true
+    private(set) var activityIndicatorButtonStates: [ActivityIndicatorButtonState] = []
+    
+    private lazy var activityIndicator: UIActivityIndicatorView = {
+        let activityIndicator = UIActivityIndicatorView()
+        activityIndicator.hidesWhenStopped = true
+        activityIndicator.color = theme?.colorStates.color(for: .selected)
```
EvansPie, line 94: The activity indicator color is light grey now (it was white before). Not sure if it makes sense to get the color for the selected state, or if it would be better to get the text color. I know that the previous behavior was just white, so can you provide the `white` color on the merchant's app example and check if it works?

EvansPie (approved):

Sources/PrimerSDK/Classes/User Interface/UIUtils.swift
```diff
@@ -0,0 +1,15 @@
...
+//  Copyright © 2022 Primer API ltd, Inc. All rights reserved.
+//
+
+import Foundation
+
+///
+/// Utility Typealias that returns a Void callback
+/// For implementation example check PrimerButton.swift
+///
+public typealias Action = () -> Void
```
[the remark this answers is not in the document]
d4r1091: Otherwise `PrimerButton:68` won't be accessible from whoever would like to implement the Button with the callback even: `public var onPressed: Action?`

Sources/PrimerSDK/Classes/Extensions & Utilities/PrimerButton.swift
d4r1091: Good point on that. The reason why I went after the selected state is that I thought from the product perspective.
The point in which we animate the button is when we press it.
So I thought that would be good for consistency to keep the activity indicator color following **the selected state**.
In case we don't have a theme though, this won't work - so I'll do the following:

`activityIndicator.color = theme?.colorStates.color(for: .selected) ?? .white`

what do you think? @EvansPie

xevious78 (approved):

PR #311, opened 2022-03-03
Release/1.28.0 beta.3

EvansPie:
DEX-1082

# What this PR does

- Adds subclasses for text fields


Example/Podfile
```diff
@@ -1,6 +1,3 @@
-#source 'https://github.com/CocoaPods/Specs.git'
-#source 'https://github.com/primer-io/primer-podspecs.git'
```
d4r1091, line 2: Is there a particular reason to remove the cocoapods main source too?

EvansPie: When you don't want to add a private repo, there's no need to add the cocoapods trunk, it defaults to this. And we don't the private repo anymore, 3DS became public.

d4r1091 (approved):

xevious78 (approved):

PR #320, opened 2022-03-16
Feature DEX-1252 Validate Token

d4r1091:
# What this PR does

### Implemented the client validation endpoint.

[Notion link](https://www.notion.so/primerio/Signing-a-Client-Token-b4a74343071c491f9d5a9ef215e5dd49#2e19fceba9994955ab942fa036ecc978)

- Reviewed the PrimerResultViewController
- Refactor `ClientService.swift` adding the validation point
- Introduced a new field for Endpoint
   - `var shouldParseResponseBody: Bool { get }` -> Default to `true`
   - To avoid custom checks based on API Verb when it comes to parsing the response
- Moved `DummySuccess` into a `typealias` of `SuccessResponse` so apparently both serves the same purpose but the latter is an actual JSON response.
- other minor refactor

If you initialize the Universal Checkout, the app should behave as normal.
In case we encounter an invalid token, at any point of the payment flow, the universal checkout should show an error.

# Before merging

_QA_

- [ x] I manually tested this with a demo application (simulator)

_Documentation_

# After merging


Example/Podfile.lock
```diff
@@ -1,7 +1,7 @@
 PODS:
-  - PrimerSDK (1.28.0-beta.4):
-    - PrimerSDK/Core (= 1.28.0-beta.4)
-  - PrimerSDK/Core (1.28.0-beta.4)
+  - PrimerSDK (1.29.2):
```
EvansPie, line 2: I released 1.29.2 this morning 😆

Please update to 1.29.3

Sources/PrimerSDK/Classes/Core/Payment Services/ClientTokenService.swift
```diff
@@ -83,35 +126,87 @@ internal class ClientTokenService: ClientTokenServiceProtocol {
...
+            }
+        }
+    }
+    
+    static func storeClientToken(_ clientToken: String, completion: @escaping (Error?) -> Void) throws {
+        
         let state: AppStateProtocol = DependencyContainer.resolve()
-        state.clientToken = nil
+        
+        // 1. Validate the token manually or return the previous one from current App State
```
EvansPie, line 158: Not entirely sure if this is ok. What's going to happen if the previous `clientToken` has been used (which means that it cannot be reused)?

Why not just reset it? The merchant provided us with an invalid token, why not just discard it?

Sources/PrimerSDK/Classes/Core/Primer/Primer.swift
```diff
@@ -244,11 +246,11 @@ public class Primer {
         Analytics.Service.record(events: [sdkEvent, connectivityEvent, timingEvent])
         
         if let clientToken = clientToken {
-            try? ClientTokenService.storeClientToken(clientToken)
+            let _ = try? ClientTokenService.storeClientToken(clientToken)
         }
         
         presentingViewController = viewController
-        show(flow: .defaultWithVault)
+        show(flow: .default)
```
EvansPie, line 253: I know naming is not good here, unfortunately that's what we have. The vault manager flow is the `defaultWithVault`, so this shouldn't be changed here since this is the `showVaultManager` function

Sources/PrimerSDK/Classes/Services/API/Primer/PrimerAPI.swift
```diff
@@ -298,6 +304,19 @@ internal extension PrimerAPI {
...
         case .fetchPayPalExternalPayerInfo(_, let payPalExternalPayerInfoRequestBody):
             return try? JSONEncoder().encode(payPalExternalPayerInfoRequestBody)
+        case .validateClientToken(_, let clientTokenToValidate):
+            return try? JSONEncoder().encode(clientTokenToValidate)
+        }
+    }
+    
+    // MARK: Should Return Response Body
+    
+    var shouldParseResponseBody: Bool {
```
EvansPie, line 314: Nice!!

EvansPie (commented):
Left a couple of comments

Example/Podfile.lock
d4r1091: at the time of writing this comment, I cannot see the new version on git 😅 
Screen Shot 2022-03-16 at 17 00 20
Screen Shot 2022-03-16 at 17 00 11

Sources/PrimerSDK/Classes/Core/Payment Services/ClientTokenService.swift
```diff
@@ -3,14 +3,19 @@
 import Foundation
 
 internal protocol ClientTokenServiceProtocol {
-    static func storeClientToken(_ clientToken: String) throws
+    static func storeClientToken(_ clientToken: String) throws -> Promise<Void>
```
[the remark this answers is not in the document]
d4r1091: I'll comment here and keep this comment open (whilst resolving the others) as they are suggesting the same thing 💪 . I thought we wanted to keep the `throws` to be consistent with the current implementation. I'll definitely update counting on Promise and completion block only. That's how I've started and then I thought I had to add the `throws` 😅 
Thanks!

d4r1091:
Closing in favour of: https://github.com/primer-io/primer-sdk-ios/pull/329

PR #341, opened 2022-04-14
Rename structures

EvansPie:
- Renames and nests payment method related structures


d4r1091 (approved):
I like the proposed renamings.
Especially after the additional context about some changes provided via Slack @EvansPie .
Awesome work

PR #364, opened 2022-05-20
DEX-1802 Remove apple pay view model if it cannot make payments

EvansPie:
DEX-1802

Remove apple pay view model if it cannot make payments


d4r1091 (approved):
👌

PR #385, opened 2022-06-13
DEX-1206 Billing Address capture

d4r1091:
# What this PR does

- This PR enables the billing address capture upon choosing to pay by card.
- Added a list of localized countries (that's the cause of ~22k lines 😅 ) 
- It is based on SDK v1 to keep supporting it
- It simplifies the codebase a bit among FormViewModels
  - All choices have been made thinking of the best compromise utilizing what we currently have
- It also integrates a new and simpler form builder which for now is used only to build the billing address itself
  - on v2 we could think of building the entire card form based on that
- general cleanup


Sources/PrimerSDK/Classes/Data Models/CountryCode.swift
```diff
@@ -525,24 +275,49 @@ internal extension CountryCode {
...
-            return "fr-BE"
-        case .es:
-            return "es-ES"
-        case .it:
-            return "it-IT"
-        default:
-            return "n/a"
+    
+    private static var languageCode: String {
+        Locale.current.languageCode ?? "en"
```
EvansPie, line 280: Where is this used? You might want to have the language code that the merchant has set first.

Sources/PrimerSDK/Classes/Extensions & Utilities/NSObject+ClassName.swift
```diff
@@ -0,0 +1,49 @@
...
+//  NSObject+ClassName.swift
+//  PrimerSDK
+//
+//  Created by Dario Carlomagno on 07/06/22.
+//
+#if canImport(UIKit)
+
+import Foundation
+
+extension NSObject {
```
EvansPie, line 11: That's a bit tricky. I believe this functionality gets exposed even if the access level is internal. It does not get exposed through the extension, but it will get exposed when someone subclasses `NSObject`.

Can you make the following test, in the Example app create a subclass of `NSObject` and check whether those vars are exposed?

If they are, try subclassing `NSObject`, extend your subclass with those vars, then use your subclass instead of `NSObject`.

Sources/PrimerSDK/Classes/Extensions & Utilities/PrimerButton.swift
```diff
@@ -179,7 +180,9 @@ extension PrimerButton {
             setImage(UIImage(), for: state)
         }
         self.activityIndicatorButtonStates = buttonStates
-        isEnabled = false
+        DispatchQueue.main.async {
+            self.isEnabled = false
```
EvansPie, line 184: I am not sure if this is needed, have you tested it? On the other hand, `activityIndicator.startAnimating()` on line 173 probably needs to be performed on the main queue.

Sources/PrimerSDK/Classes/Extensions & Utilities/PrimerButton.swift
```diff
@@ -188,7 +191,9 @@ extension PrimerButton {
             setTitle(buttonState.title, for: buttonState.state)
             setImage(buttonState.image, for: buttonState.state)
         }
-        isEnabled = true
+        DispatchQueue.main.async {
+            self.isEnabled = true
```
EvansPie, line 195: Same as above

Sources/PrimerSDK/Classes/Extensions & Utilities/StringExtension.swift
```diff
@@ -169,16 +169,23 @@ internal extension String {
...
     }
     
-    var isValidCardholderName: Bool {
+    var isValidNonDecimalString: Bool {
         if isEmpty { return false }
         return rangeOfCharacter(from: .decimalDigits) == nil
     }
     
+    var isValidCardholderName: Bool {
+        return isValidNonDecimalString
```
EvansPie, line 182: Can you review those names and make them more declarative? Or if they are declarative, can you review the `isValidCardholderName` functionality?

I would expect `isValidNonDecimalString` to allow `44o!` to pass, but this is not a valid cardholder name.

Sources/PrimerSDK/Classes/User Interface/Components/PrimerFormView.swift
```diff
@@ -0,0 +1,95 @@
+//
+//  PrimerFormView.swift
+//  PrimerSDK
+//
+//  Created by Dario Carlomagno on 09/06/22.
+//
+
+import Foundation
```
EvansPie, line 8: You need `#if canImport(UIKit)` and also `import UIKit` since you're dealing with views.

Sources/PrimerSDK/Classes/User Interface/Countries/CountrySelectorViewController.swift
```diff
@@ -0,0 +1,77 @@
...
+                                                bundle: Bundle.primerResources,
+                                                value: "Choose country",
+                                                comment: "Choose country - Choose your billing address country")
+        bankTitleLabel.font = UIFont.systemFont(ofSize: 20)
+        bankTitleLabel.textColor = theme.text.title.color
+        verticalStackView.addArrangedSubview(bankTitleLabel)
+        
+        verticalStackView.addArrangedSubview(viewModel.searchCountryTextField)
+                
+        let separator2 = UIView()
```
EvansPie, line 67: Where's separator 1? 😛

Sources/PrimerSDK/Classes/User Interface/Text Fields/PrimerCardNumberFieldView.swift
```diff
@@ -18,48 +18,23 @@ public final class PrimerCardNumberFieldView: PrimerTextFieldView {
...
-                    issuerId: nil,
-                    paymentMethodType: PaymentMethodConfigType.paymentCard.rawValue,
-                    url: nil),
-                extra: nil,
-                objectType: .input,
-                objectId: .cardNumber,
-                objectClass: "\(Self.self)",
-                place: .cardForm))
-        Analytics.Service.record(event: viewEvent)
+        super.textFieldDidBeginEditing(textField)
```
EvansPie, line 33: Why not remove the function?

Sources/PrimerSDK/Classes/User Interface/Text Fields/PrimerTextFieldView+Analytics.swift
```diff
@@ -0,0 +1,44 @@
+//
+//  PrimerTextFieldView+Analytics.swift
+//  PrimerSDK
+//
+//  Created by Dario Carlomagno on 07/06/22.
+//
+
+#if canImport(UIKit)
+
+import Foundation
```
EvansPie, line 10: Import `UIKit`. Please check all the new files for this.

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/FormsTokenizationViewModel/CardFormPaymentMethodTokenizationViewModel.swift
```diff
@@ -749,7 +870,7 @@ extension CardFormPaymentMethodTokenizationViewModel: CardComponentsManagerDeleg
             } else if primerTextFieldView is PrimerCardholderNameFieldView {
                 cardholderNameContainerView?.errorText = "Invalid name"
             } else if primerTextFieldView is PrimerPostalCodeFieldView {
-                postalCodeContainerView.errorText = "\(localPostalCodeTitle) is required" // todo: localise if UK, etc.
+                postalCodeContainerView.errorText = "\(PrimerPostalCodeField.localPostalCodeTitle) is required" // todo: localise if UK, etc.
```
EvansPie, line 873: Can you make sure this gets localized?

EvansPie (changes requested):

Sources/PrimerSDK/Classes/Extensions & Utilities/PrimerButton.swift
d4r1091: I've tested it yeah. Everything is actually performed on the main thread so that `DispatchQueue.main.async {` block shouldn't even be needed. Although in order to trigger the expected animation upon setting the `PrimerButton` disabled, that's the only solution I found to workaround the isssue.

Sources/PrimerSDK/Classes/Extensions & Utilities/PrimerButton.swift
d4r1091: https://github.com/primer-io/primer-sdk-ios/pull/385#discussion_r896662664

Sources/PrimerSDK/Classes/User Interface/Countries/CountrySelectorViewController.swift
d4r1091: 🤦 solved

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/FormsTokenizationViewModel/CardFormPaymentMethodTokenizationViewModel.swift
d4r1091: Solved ✅

Sources/PrimerSDK/Classes/Extensions & Utilities/NSObject+ClassName.swift
d4r1091: Thanks for the inputs 💪 - checked. I believe we are good. (Screenshot from a swift file within the example app)

Screen Shot 2022-06-14 at 17 21 58

Sources/PrimerSDK/Classes/Extensions & Utilities/StringExtension.swift
d4r1091: I reviewed the logic from within the:

```
func textField(_ textField: UITextField, shouldChangeCharactersIn range: NSRange, replacementString string: String) -> Bool {
```

of each FieldView where it is appropriate to make a better check as we type.
The `PrimerSimpleCardFormTextFieldView` will still allow decimal inputs.

Sources/PrimerSDK/Classes/User Interface/Text Fields/PrimerCardNumberFieldView.swift
d4r1091: ✅ 🙏

Sources/PrimerSDK/Classes/Data Models/CountryCode.swift
EvansPie: Yeah there is, via the `PrimerSettings`. But it doesn't apply on this case.

EvansPie (approved):

d4r1091:
Closing in favor of https://github.com/primer-io/primer-sdk-ios/pull/398

PR #397, opened 2022-06-20
DEX 1968 remove flow

EvansPie:
DEX-1968

# What this PR does

- Remove PrimerSessionFlow as part of the SDK v2 internal refactor


Sources/PrimerSDK/Classes/Data Models/PaymentMethodConfigurationType.swift
```diff
@@ -251,30 +251,26 @@ public enum PrimerPaymentMethodType: Codable, Equatable, Hashable {
                 .payNLGiropay,
                 .payNLIdeal,
                 .payNLPayconiq,
+                .primerTestKlarna,
+                .primerTestPayPal,
```
d4r1091, line 255: I thought I put `.primerTestKlarna` and `.primerTestPaypal` in the right place the first time.
Shouldn’t they go at the bottom (`return true`) instead of being among the ones where we check the `Primer.shared.intent == .checkout` ?

d4r1091:
Closing as it would come as part of: https://github.com/primer-io/primer-sdk-ios/pull/403

PR #401, opened 2022-06-29
Release 1.36.2

EvansPie:
Release 1.36.2

- Fix Adyen iDeal bank selector for v1

### DIFF
--- a/Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckout.swift
+++ b/Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckout.swift
@@ -251,7 +251,6 @@ public class PrimerHeadlessUniversalCheckout {
     
     public func showPaymentMethod(_ paymentMethod: PaymentMethodConfigType) {
         DispatchQueue.main.async {
-            
             var settings: PrimerSettingsProtocol = DependencyContainer.resolve()
             settings.hasDisabledSuccessScreen = true
             settings.isInitialLoadingHidden = true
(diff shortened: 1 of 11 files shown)


Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckout.swift
```diff
@@ -25,13 +25,8 @@ public class PrimerHeadlessUniversalCheckout {
...
         
-        guard PrimerHeadlessUniversalCheckout.current.delegate != nil else {
-            let err = PrimerError.missingPrimerCheckoutComponentsDelegate(userInfo: ["file": #file, "class": "\(Self.self)", "function": #function, "line": "\(#line)"])
-            ErrorHandler.handle(error: err)
-            DispatchQueue.main.async {
-                completion(nil, err)
-            }
-            return
+        if PrimerHeadlessUniversalCheckout.current.delegate == nil {
+            print("WARNING: PrimerHeadlessUniversalCheckout delegate has not been set, and you won't be able to receive the Payment Method Token data to create a payment.")
```
d4r1091, line 29: Is there a specific reason for raising the Warning and not the ErrorHandler? I believe e I asked before but just wanted a 🧠 refresh. Especially as a similar  `print` is added in another change alongside the ErrorHandler

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/BankSelectorTokenizationViewModel.swift
```diff
@@ -193,12 +193,20 @@ class BankSelectorTokenizationViewModel: ExternalPaymentMethodTokenizationViewMo
         
     @objc
     override func startTokenizationFlow() {
-        super.startTokenizationFlow()
```
d4r1091, line 196: Don't we need `super` here anymore?

Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckout.swift
EvansPie: We don't want to break the flow if someone doesn't use the HUC, but somehow calls something from the HUC

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/BankSelectorTokenizationViewModel.swift
EvansPie: We need to override cause there're a couple of steps that have to be done before the steps that are done in `super`.

PR #452, opened 2022-08-22
DEX-2276 OVO via Xendit HUC Raw Data

d4r1091:
# What this PR does

- Implemented OVO via Xendit
- Reviewed RawCardDataManager now decoupled from handling Card only RawData
- Minor bugfixes

### DIFF
--- a/Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckoutProtocols.swift
+++ b/Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckoutProtocols.swift
@@ -154,7 +154,7 @@ public enum PrimerInputElementType: Int {
                 .cvv,
                 .otp,
                 .phoneNumber:
-            return CharacterSet.decimalDigits
+            return CharacterSet(charactersIn: "0123456789")
             
         case .cardholderName:
             return CharacterSet.letters
--- a/Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckoutProtocols.swift
+++ b/Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckoutProtocols.swift
@@ -13,7 +13,14 @@ import UIKit
 @objc
 public enum PrimerInputElementType: Int {
     
-    case cardNumber, expiryDate, cvv, cardholderName, otp, postalCode, unknown
+    case cardNumber
+    case expiryDate
+    case cvv
+    case cardholderName
+    case otp
+    case postalCode
+    case phoneNumber
+    case unknown
     
     public var stringValue: String {
         switch self {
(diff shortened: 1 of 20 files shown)


Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckoutProtocols.swift
```diff
@@ -136,46 +149,35 @@ public enum PrimerInputElementType: Int {
...
-            return CharacterSet(charactersIn: "0123456789")
-            
-        case .cvv:
-            return CharacterSet(charactersIn: "0123456789")
+        case .cardNumber,
+                .expiryDate,
+                .cvv,
+                .otp,
+                .phoneNumber:
+            return CharacterSet.decimalDigits
```
EvansPie, line 157: I am not very sure about the `decimalDigits`. What does it include? Does it change in other writing systems? Why not using the strictly defined set `CharacterSet(charactersIn: "0123456789")`?

Sources/PrimerSDK/Classes/Extensions & Utilities/StringExtension.swift
```diff
@@ -209,6 +209,12 @@ internal extension String {
         let emailP = NSPredicate(format: "SELF MATCHES %@", emailRegEx)
         return emailP.evaluate(with: self)
     }
+    
+    var isValidPhoneNumber: Bool {
+        let phoneNumberRegEx = "^(^\\+628|628)(\\d{8,10})"
```
EvansPie, line 214: Have you aligned with Android on the regex?

Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckoutProtocols.swift
d4r1091: Yeah `decimalDigits` also includes representations like the Arabic numeral ones. Do you think it's better to revert that to `0123456789` ? I thought that if we expect decimal numbers strictly being `0123456789` format, we should explain that to the user.

Sources/PrimerSDK/Classes/Extensions & Utilities/StringExtension.swift
d4r1091: Yes - aligned with Android on that, we can keep this one very specific to OVO and then expand to have different validations for other APMs

Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerRawDataManager.swift
```diff
@@ -16,81 +16,76 @@ public protocol PrimerRawDataManagerDelegate {
...
+    func validateRawData(_ data: PrimerRawData) -> Promise<Void>
+}
+
 extension PrimerHeadlessUniversalCheckout {
     
     public class RawDataManager: NSObject {
         
         public var delegate: PrimerRawDataManagerDelegate?
         public private(set) var paymentMethodType: String
-        public private(set) var requiredInputElementTypes: [PrimerInputElementType]
```
EvansPie, line 25: That's already in our public facing API

EvansPie (approved):

PR #454, opened 2022-08-25
Feature DEX 2335 klarna in app

EvansPie:
DEX-2335

# What this PR does

Replace Klarna's webview with the Klarna mobile SDK


Example/PrimerSDK/Info.plist
```diff
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>NSCameraUsageDescription</key>
+	<string>This app needs camera access to scan cards</string>
```
d4r1091, line 6: Are we going to document the need for this permission for whoever will be integrating our SDK?

Example/PrimerSDK/Info.plist
```diff
@@ -36,6 +38,8 @@
 	<key>LSApplicationQueriesSchemes</key>
 	<array>
 		<string>bankid</string>
+		<string>klarna</string>
+		<string>klarnaconsent</string>
```
d4r1091, line 42: Same for these two

Sources/PrimerSDK/Classes/Core/Primer/Primer.swift
```diff
@@ -1,9 +1,9 @@
 #if canImport(UIKit)
+import UIKit
```
d4r1091, line 2: good catch 👌

Sources/PrimerSDK/Classes/Data Models/PrimerConfiguration.swift
```diff
@@ -37,6 +37,13 @@ struct PrimerAPIConfiguration: Codable {
             }
         }
         
+        #if !canImport(PrimerKlarnaSDK)
+        if let klarnaViewModelIndex = viewModels.firstIndex(where: { $0.config.type == PrimerPaymentMethodType.klarna.rawValue }) {
+            viewModels.remove(at: klarnaViewModelIndex)
+            print("WARNING!\nKlarna configuration has been found but module 'Primer/Klarna' is missing. Add `Primer/Klarna' in your project so you can perform payments with Klarna.")
```
d4r1091, line 43: I guess the copy **might** be misleading. It seems like when we say `Primer/Klarna` we refer to a submodule of our PrimerSDK, while as far as I can see, it's an entirely different SDK, similar to Primer3DS.
I've also seen that we refer to it as a "module".
Considering the above my suggestion would be something like this:
```suggestion
            print("WARNING!\nKlarna configuration has been found, but the `PrimerKlarnaSDK` module is missing. Add `PrimerKlarnaSDK` via Pods or SwiftPackageManager in your project so you can perform payments with Klarna.")
```

Sources/PrimerSDK/Classes/Error Handler/PrimerError.swift
```diff
@@ -338,6 +338,7 @@ internal enum PrimerError: PrimerErrorProtocol {
     case cancelledByCustomer(message: String?, userInfo: [String: String]?, diagnosticsId: String?)
     case paymentFailed(userInfo: [String: String]?, diagnosticsId: String?)
     case applePayTimedOut(userInfo: [String: String]?, diagnosticsId: String?)
+    case failedToFindModule(name: String, userInfo: [String: String]?, diagnosticsId: String?)
```
d4r1091, line 341: Will we raise this if we fail to import 3DS too? Perhaps in the future

Sources/PrimerSDK/Classes/Error Handler/PrimerError.swift
```diff
@@ -664,6 +672,8 @@ internal enum PrimerError: PrimerErrorProtocol {
             return nil
         case .applePayTimedOut:
             return "Make sure you have an active internet connection and your Apple Pay configuration is correct."
+        case .failedToFindModule(let name, _, _):
+            return "Make sure you have added module \(name) in your project"
```
d4r1091, line 676:
```suggestion
            return "Make sure you have added the module \(name) in your project."
```

d4r1091 (commented):
Left a couple of suggestions 👌

Sources/PrimerSDK/Classes/Core/Primer/Primer.swift
```diff
@@ -152,44 +152,12 @@ public class Primer {
...
-            let err = PrimerError.unsupportedIntent(
-                intent: intent,
-                userInfo: ["file": #file, "class": "\(Self.self)", "function": #function, "line": "\(#line)"],
-                diagnosticsId: nil)
-            ErrorHandler.handle(error: err)
-            PrimerDelegateProxy.raisePrimerDidFailWithError(err, data: nil)
-            completion?(err)
-            return
-        }
-
```
d4r1091, line 186: Isn't this needed anymore?

Example/PrimerSDK/Info.plist
EvansPie: Yup, it's only needed if `PrimerKlarnaSDK` is installed though

Example/PrimerSDK/Info.plist
EvansPie: As above

PR #477, opened 2022-09-09
New readme

dirkbrnd:
DEX-2317

# What this PR does

Adds a Readme!


README.md
```diff
@@ -1,7 +1,138 @@
...
+import PrimerSDK
+ 
+class MyViewController: UIViewController {
+ 
+    override func viewDidLoad() {
+        super.viewDidLoad()
+ 
+        // Initialize the SDK with the default settings.
+        let settings = PrimerSettings()
+        Primer.shared.configure(settings: settings, delegate: self)
```
d4r1091, line 102: To configure Primer with default `PrimerSettings` we can omit the argument passed to the configuration function.
If the intention was to show the PrimerSetting object, fair enough 👍.
Otherwise, we can apply the suggestion below:
```suggestion
        Primer.shared.configure(delegate: self)
```

d4r1091 (commented):
A solid starting point 🙌

d4r1091 (approved):

PR #485, opened 2022-09-19
DEVX-205 Replace Klarna redirect URL

EvansPie:
DEVX-205

Replace `hppRedirectUrl` with `urlScheme`


Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/KlarnaTokenizationViewModel.swift
```diff
@@ -346,15 +357,6 @@ class KlarnaTokenizationViewModel: PaymentMethodTokenizationViewModel {
                         function: #function
                     )
                     
-                    self?.klarnaPaymentSession = res
```
d4r1091, line 349: don't we need this anymore?

EvansPie: Good catch, but it's alright as it's already done on line 117

d4r1091: even better 👌

d4r1091 (approved):

PR #505, opened 2022-09-26
DEVX-111

EvansPie:
DEVX-111

- Refactors delegate
- Renames `DecodedClientToken` to `DecodedJWTToken`
- Sets client token and api configuration via `PrimerAPIConfigurationModule`
- Injects `apiClient` through static var (will be refactored when we have mocks for all modules)
- Replaces local functions with polling module on HUC
- Moves `exchangeToken` in `TokenizationService`
- Moves `ClientSessionService` and `ClientTokenService` logic in `PrimerAPIConfigurationModule`
- Adds unit tests for the drop-in UI
- Adds unit tests for the polling module
- Adds unit tests for the api configuration module
- Adds unit tests for the user interface module


dirkbrnd (approved):

d4r1091 (commented):
The PR and the unit tests are absolute 👌 🚀 .
I wonder whether we should keep all the commented code or remove it.
Technically, we would need to rewrite those too.
I've also fixed an issue on the Danger folder matching the correct UnitTests one 💪 as thanks to this PR I've realised that the warning is definitely not appropriate here 😅

d4r1091 (approved):
Discussed to leave the commented tests 👍 So will approve

PR #509, opened 2022-09-28
CHKT-21 Bancontact debit via adyen

d4r1091:
# What this PR does

- Added Adyen Bancontact UI and general payment method parsing
- Reviewed `PrimerButton`'s code in `UserInterfaceModule`
- Reviewed Polling module for Card Tokenization View model
- Added support for multiple payment methods on Headless Universal Checkout UI Manager
  - So far, only `.paymentCard` has been supported
- Implemented Adyen Bancontact in Headless Raw Card
  - Implemented a generic `PrimerCardRedirectData` object to handle similar scenarios in the future
- Marked initializer of `PrimerHeadlessUniversalCheckoutUIManager.swift` as deprecated in favor of the one accepting multiple payment methods.
(I know we'd like to mark as `@deprecated` multiple chunks of code at once, so let's skip it if not appropriate this time.


Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckoutUIManager.swift
```diff
@@ -109,14 +109,21 @@ extension PrimerHeadlessUniversalCheckout {
             log(logLevel: .debug, message: "🧨 deinit: \(self) \(Unmanaged.passUnretained(self).toOpaque())")
         }
         
+        public init(paymentMethodType: String) {
+            self.paymentMethodType = paymentMethodType
```
EvansPie, line 113: This requires a check, otherwise what's going to happen if it gets initialized with an unsupported type?

Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckoutUIManager.swift
```diff
@@ -184,68 +191,122 @@ extension PrimerHeadlessUniversalCheckout {
         }
         
         private func buildRequestBody() -> Promise<Request.Body.Tokenization> {
+            switch self.paymentMethodType {
+            case PrimerPaymentMethodType.paymentCard.rawValue:
+                return makeCardRequestBody()
+            case PrimerPaymentMethodType.adyenBancontact.rawValue:
+                return makeCardRedirectRequestBody()
+            default:
+                fatalError()
```
EvansPie, line 200: Please do this a `precondition` and return an error to the delegate

Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckoutUIManager.swift
```diff
@@ -489,7 +550,7 @@ extension PrimerHeadlessUniversalCheckout {
             }
         }
         
-        private func presentWeb3DS(with redirectUrl: URL) -> Promise<Void> {
+        private func presentWebRedirectViewControllerWithRedirectUrl(_ redirectUrl: URL) -> Promise<Void> {
```
EvansPie, line 553: Suggestion: Maybe this could be renamed to the more generic `presentUI` that we're using on the view models

Sources/PrimerSDK/Classes/Data Models/PrimerPaymentMethodType.swift
```diff
@@ -10,6 +10,7 @@ internal enum PrimerPaymentMethodType: String, Codable, CaseIterable, Equatable,
     
     case adyenAlipay        = "ADYEN_ALIPAY"
     case adyenBlik          = "ADYEN_BLIK"
+    case adyenBancontact    = "ADYEN_BANCONTACT_CARD"
```
EvansPie, line 13: Can you check whether there's a chance to have `adyenBancontact` in the future? If that's the case, then it would be better to rename `adyenBancontact` to `adyenBancontactCard`

Sources/PrimerSDK/Classes/PCI/Checkout Components/PrimerCardData.swift
```diff
@@ -0,0 +1,69 @@
...
+//  PrimerSDK
+//
+//  Created by Dario Carlomagno on 27/09/22.
+//
+
+#if canImport(UIKit)
+
+import Foundation
+
+public class PrimerCardData: PrimerRawData {
```
EvansPie, line 12: I don't think that has been modified at all, just moved, right?

Sources/PrimerSDK/Classes/PCI/Checkout Components/PrimerCardRedirectData.swift
```diff
@@ -0,0 +1,63 @@
...
+//  Created by Dario Carlomagno on 27/09/22.
+//
+
+#if canImport(UIKit)
+
+import Foundation
+
+public class PrimerCardRedirectData: PrimerRawData {}
+
+public class PrimerBancontactCardRedirectData: PrimerCardRedirectData {
```
EvansPie, line 14: This could be extended by `PrimerCardData`, then `PrimerCardData` would only have the cvv. This is just a suggestion for the future.

Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckoutUIManager.swift
d4r1091: I've made the function to `throw` an error in case of an unsupported payment method type so that the chain is interrupted

Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckoutUIManager.swift
d4r1091: ✅

Sources/PrimerSDK/Classes/PCI/Checkout Components/PrimerCardData.swift
d4r1091: Correct

Sources/PrimerSDK/Classes/PCI/Checkout Components/PrimerCardRedirectData.swift
d4r1091: 🙏 yeah I get the idea - thanks!

Sources/PrimerSDK/Classes/Data Models/PrimerPaymentMethodType.swift
d4r1091: Good one 👍 - updated as I've seen there were discussions about this naming. Better to have an explicit one, matching the classes' names we have for the CardData

EvansPie (approved):

PR #527, opened 2022-10-14
Feature DEVX-355

EvansPie:
DEVX-355

Refactor example app


d4r1091:
Released as part of https://github.com/primer-io/primer-sdk-ios/pull/529

PR #541, opened 2022-11-03
Testing tags and PR management

Tino-Primer:
Testing CI


Tino-Primer:
I didn't add a @Regression tag, instead we run all of our existing tests when trying to merge to Master and only a few selected @Smoke tests for other PRs, which I assumed would be a similar approach. 
If you guys want to have a @Regression tag I'll add it.

PR #559, opened 2022-12-19
Release 2.16.0

EvansPie:
Release 2.16.0

- Integrate iPay88
- Add Headless analytics events
- Add properties in analytics events
- Return error on Apple Pay on simulator

### DIFF
--- a/Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckout.swift
+++ b/Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckout.swift
@@ -184,7 +184,7 @@ public class PrimerHeadlessUniversalCheckout {
 #endif
         
 #if !canImport(PrimerIPay88SDK)
-        if let iPay88ViewModelIndex = paymentMethods.firstIndex(where: { $0.type == PrimerPaymentMethodType.iPay88Card.rawValue }) {
+        if let iPay88ViewModelIndex = paymentMethods?.firstIndex(where: { $0.type == PrimerPaymentMethodType.iPay88Card.rawValue }) {
             paymentMethods.remove(at: iPay88ViewModelIndex)
             print("\nWARNING!\niPay88 configuration has been found but module 'PrimerIPay88SDK' is missing. Add `PrimerIPay88SDK' in your project by adding \"pod 'PrimerIPay88SDK'\" in your podfile or by adding \"primer-ipay88-sdk-ios\" in your Swift Package Manager, so you can perform payments with iPay88.\n\n")
         }
--- a/Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckout.swift
+++ b/Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckout.swift
@@ -185,7 +185,7 @@ public class PrimerHeadlessUniversalCheckout {
         
 #if !canImport(PrimerIPay88SDK)
         if let iPay88ViewModelIndex = paymentMethods?.firstIndex(where: { $0.type == PrimerPaymentMethodType.iPay88Card.rawValue }) {
-            paymentMethods.remove(at: iPay88ViewModelIndex)
+            paymentMethods?.remove(at: iPay88ViewModelIndex)
             print("\nWARNING!\niPay88 configuration has been found but module 'PrimerIPay88SDK' is missing. Add `PrimerIPay88SDK' in your project by adding \"pod 'PrimerIPay88SDK'\" in your podfile or by adding \"primer-ipay88-sdk-ios\" in your Swift Package Manager, so you can perform payments with iPay88.\n\n")
         }
 #endif
--- a/Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckout.swift
+++ b/Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckout.swift
@@ -174,12 +174,22 @@ public class PrimerHeadlessUniversalCheckout {
     
     internal func listAvailablePaymentMethodsTypes() -> [String]? {
         var paymentMethods = PrimerAPIConfiguration.paymentMethodConfigs
-        if let klarnaIndex = paymentMethods?.firstIndex(where: { $0.type == PrimerPaymentMethodType.klarna.rawValue }) {
+        
 #if !canImport(PrimerKlarnaSDK)
+        if let klarnaIndex = paymentMethods?.firstIndex(where: { $0.type == PrimerPaymentMethodType.klarna.rawValue }) {
+
             paymentMethods?.remove(at: klarnaIndex)
             print("\nWARNING!\nKlarna configuration has been found but module 'PrimerKlarnaSDK' is missing. Add `PrimerKlarnaSDK' in your project by adding \"pod 'PrimerKlarnaSDK'\" in your podfile or by adding \"primer-klarna-sdk-ios\" in your Swift Package Manager, so you can perform payments with Klarna.\n\n")
-#endif
         }
+#endif
+        
+#if !canImport(PrimerIPay88SDK)
+        if let iPay88ViewModelIndex = paymentMethods.firstIndex(where: { $0.type == PrimerPaymentMethodType.iPay88Card.rawValue }) {
+            paymentMethods.remove(at: iPay88ViewModelIndex)
+            print("\nWARNING!\niPay88 configuration has been found but module 'PrimerIPay88SDK' is missing. Add `PrimerIPay88SDK' in your project by adding \"pod 'PrimerIPay88SDK'\" in your podfile or by adding \"primer-ipay88-sdk-ios\" in your Swift Package Manager, so you can perform payments with iPay88.\n\n")
+        }
+#endif
+        
         return paymentMethods?.compactMap({ $0.type }).filter({ !unsupportedPaymentMethodTypes.contains($0) })
     }
     
--- a/Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckout.swift
+++ b/Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckout.swift
@@ -31,7 +31,22 @@ public class PrimerHeadlessUniversalCheckout {
     
     fileprivate init() {}
     
-    public func start(withClientToken clientToken: String, settings: PrimerSettings? = nil, delegate: PrimerHeadlessUniversalCheckoutDelegate? = nil, completion: @escaping (_ paymentMethodTypes: [String]?, _ err: Error?) -> Void) {
+    public func start(
+        withClientToken clientToken: String,
+        settings: PrimerSettings? = nil,
+        delegate: PrimerHeadlessUniversalCheckoutDelegate? = nil,
+        completion: @escaping (_ paymentMethodTypes: [String]?, _ err: Error?) -> Void
+    ) {
+        DependencyContainer.register(settings ?? PrimerSettings() as PrimerSettingsProtocol)
+        
+        if delegate != nil {
+            PrimerHeadlessUniversalCheckout.current.delegate = delegate
+        }
+        
+        if PrimerHeadlessUniversalCheckout.current.delegate == nil {
+            print("WARNING!\nPrimerHeadlessUniversalCheckout delegate has not been set, and you won't be able to receive the Payment Method Token data to create a payment.")
+        }
+        
         PrimerInternal.shared.sdkIntegrationType = .headless
         PrimerInternal.shared.intent = .checkout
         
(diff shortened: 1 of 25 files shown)


Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/ApplePayTokenizationViewModel.swift
```diff
@@ -397,6 +397,24 @@ extension ApplePayTokenizationViewModel: PKPaymentAuthorizationViewControllerDel
         didAuthorizePayment payment: PKPayment,
         handler completion: @escaping (PKPaymentAuthorizationResult) -> Void
     ) {
+#if targetEnvironment(simulator)
+        if payment.token.paymentData.count == 0 {
+            let err = PrimerError.invalidArchitecture(
+                description: "Apple Pay cannot return tokenization data in the simulator",
```
xevious78, line 403: Merchants don't know what "tokenization data" means here.
I'd put sth like: "Apple Pay does not work with Primer when used in the simulator due to a limitation from Apple Pay."

Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckout.swift
```diff
@@ -39,7 +47,28 @@ public class PrimerHeadlessUniversalCheckout {
...
+        
+        self.checkoutSessionId = UUID().uuidString
+        self.timingEventId = UUID().uuidString
+        
+        let sdkEvent = Analytics.Event(
+            eventType: .sdkEvent,
+            properties: SDKEventProperties(
+                name: #function,
+                params: [
+                    "intent": PrimerInternal.shared.intent?.rawValue ?? "null"
```
xevious78, line 61: Any way we can pass al parameters? `clientToken` and `settings` are also valuable?

Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckout.swift
```diff
@@ -181,16 +231,43 @@ public class PrimerHeadlessUniversalCheckout {
     }
     
     public static func makeButton(for paymentMethodType: String) -> UIButton? {
+        let sdkEvent = Analytics.Event(
+            eventType: .sdkEvent,
+            properties: SDKEventProperties(
+                name: #function,
+                params: nil))
```
xevious78, line 238: Can we have full visibility on the args used?

Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckout.swift
```diff
@@ -39,7 +47,28 @@ public class PrimerHeadlessUniversalCheckout {
...
+                ]))
+        
+        let timingStartEvent = Analytics.Event(
+            eventType: .timerEvent,
+            properties: TimerEventProperties(
+                momentType: .start,
+                id: self.timingEventId))
+        
+        Analytics.Service.record(events: [sdkEvent, timingStartEvent])
+        Analytics.Service.sync()
```
xevious78, line 71: Why do you need to call `sync` all the time?

Sources/PrimerSDK/Classes/Data Models/ClientToken.swift
```diff
@@ -28,6 +28,10 @@ struct DecodedJWTToken: Codable {
     var qrCode: String?
     var accountNumber: String?
     
+    // iPay88
+    var backendCallbackUrl: String?
+    var primerTransactionId: String?
```
xevious78, line 33: Is there any plan to split this file into multiple files? It seems that having only one file for all possibilities of the client token is a bit hard to follow 🤔

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/IPay88TokenizationViewModel.swift
```diff
@@ -0,0 +1,547 @@
...
+    private var didCancel: (() -> Void)?
+    private var didComplete: (() -> Void)?
+    private var didFail: ((_ err: PrimerError) -> Void)?
+#endif
+    
+    private lazy var iPay88NumberFormatter: NumberFormatter = {
+        let numberFormatter = NumberFormatter()
+        numberFormatter.numberStyle = .decimal
+        numberFormatter.decimalSeparator = "."
+        numberFormatter.groupingSeparator = ","
```
xevious78, line 39: Are we 100% sure this is not tied to any locale?

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/IPay88TokenizationViewModel.swift
```diff
@@ -0,0 +1,547 @@
...
+        numberFormatter.maximumFractionDigits = 2
+        numberFormatter.minimumFractionDigits = 2
+        return numberFormatter
+    }()
+    
+    override func validate() throws {
+        guard let decodedJWTToken = PrimerAPIConfigurationModule.decodedJWTToken, decodedJWTToken.isValid else {
+            let err = PrimerError.invalidClientToken(
+                userInfo: ["file": #file, "class": "\(Self.self)", "function": #function, "line": "\(#line)"],
```
xevious78, line 48: I'm seeing you using this everywhere. I understand you cannot use another function cuz `#file` and `#function` would point to another place. 
But could use preprocessing?

```
#define ANALYTICS_DEFAULT_USER_INFO ["file": #file, "class": "\(Self.self)", "function": #function, "line": "\(#line)"]
```

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/IPay88TokenizationViewModel.swift
```diff
@@ -0,0 +1,547 @@
...
+        
+        if AppState.current.currency != .MYR {
+            let err = PrimerError.invalidClientSessionValue(
+                name: "currency",
+                value: AppState.current.currency?.rawValue,
+                userInfo: ["file": #file, "class": "\(Self.self)", "function": #function, "line": "\(#line)"],
+                diagnosticsId: nil)
+            ErrorHandler.handle(error: err)
+            errors.append(err)
+        }
```
xevious78, line 104: These checks should be done before showing the payment method button to the user / passing it to `onAvailablePaymentMethodsLoad`.

No need to show a payment method if know it won't work

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/IPay88TokenizationViewModel.swift
```diff
@@ -0,0 +1,547 @@
...
+            ErrorHandler.handle(error: err)
+            throw err
+        }
+    }
+    
+    override func performPreTokenizationSteps() -> Promise<Void> {
+        let event = Analytics.Event(
+            eventType: .ui,
+            properties: UIEventProperties(
+                action: .click,
```
xevious78, line 219: Don't think that's the right place to call this.
We would be logging a `.click` event in headless checkout, which doesn't make much sense here 🤔

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/IPay88TokenizationViewModel.swift
```diff
@@ -0,0 +1,547 @@
...
+                seal.fulfill()
+            }
+            .ensure {
+
+            }
+            .catch { err in
+                seal.reject(err)
+            }
+            
+#else
```
xevious78, line 255: Same thing here: we shouldn't return `iPay88` in the list of payment methods if the SDK is missing

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/IPay88TokenizationViewModel.swift
```diff
@@ -0,0 +1,547 @@
...
+                    self.primerIPay88ViewController.isModalInPresentation = true
+                    self.primerIPay88ViewController.modalPresentationStyle = .fullScreen
+                }
+
+                self.willPresentPaymentMethodUI?()
+                PrimerUIManager.primerRootViewController?.present(self.primerIPay88ViewController, animated: true, completion: {
+                    DispatchQueue.main.async {
+                        PrimerHeadlessUniversalCheckout.current.delegate?.primerHeadlessUniversalCheckoutPaymentMethodDidShow?(for: self.config.type)
+                        self.didPresentPaymentMethodUI?()
+                        seal.fulfill(())
```
xevious78, line 449: What does `fulfill(())` means? (empty parathensis)

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/IPay88TokenizationViewModel.swift
```diff
@@ -0,0 +1,547 @@
...
+            self.didCancel = {
+                let err = PrimerError.cancelled(
+                    paymentMethodType: self.config.type,
+                    userInfo: ["file": #file, "class": "\(Self.self)", "function": #function, "line": "\(#line)"],
+                    diagnosticsId: nil)
+                ErrorHandler.handle(error: err)
+                pollingModule.cancel(withError: err)
+            }
+            
+            self.didFail = { err in
```
xevious78, line 477: If I get this right: this automatically outputs a log to Primer?

xevious78 (approved):
Looking good. Got some comments

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/ApplePayTokenizationViewModel.swift
EvansPie: Fixed

Sources/PrimerSDK/Classes/Data Models/ClientToken.swift
EvansPie: Yeah, there is once the `clientToken` is refactored.

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/IPay88TokenizationViewModel.swift
EvansPie: Isn't this supposed to be driven by the BE? If a PM is not valid for a PM, then BE shouldn't return it in the response, and therefore not rendered. Right?

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/IPay88TokenizationViewModel.swift
EvansPie: Yeah, you're right

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/IPay88TokenizationViewModel.swift
EvansPie: Same as fulfilling with `Void`

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/IPay88TokenizationViewModel.swift
EvansPie: `iPay88` would never be here, because the view model will be removed from the `PrimerAPIConfiguration.paymentMethodConfigViewModels` if `PrimerIPay88SDK` is missing.

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/IPay88TokenizationViewModel.swift
EvansPie: `iPay88` has no docs on this, I tried with a couple of combinations. Never managed to get the `groupingSeparator` cause they don't allow payments in sandbox over 999.99, they throw an `Transaction overlimit` error. @semirp anything on Android?

Sources/PrimerSDK/Classes/Core/Checkout Components/PrimerHeadlessUniversalCheckout.swift
EvansPie: No need actually. Removed them.

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/IPay88TokenizationViewModel.swift
EvansPie: +1

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/IPay88TokenizationViewModel.swift
EvansPie: If it cancels on HUC, an `cancellation` error will be returned. If it's dropin, it will return on the dropin UI.

If it's an error, it will be passed via the delegates.

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/IPay88TokenizationViewModel.swift
semirp: They just say we should use 1,234.55 format @EvansPie

Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/IPay88TokenizationViewModel.swift
semirp: @EvansPie similar to what we did to other APMs. We verify it **before** and log warnings if something is wrong. Probably you did it for Apple Pay

PR #571, opened 2023-02-02
CHKT-1013 modify errors and analytics events

EvansPie:
CHKT-1013

- Fix error's properties on analytics events
- Add 3DS import error

### DIFF
--- a/Sources/PrimerSDK/Classes/Core/Analytics/Analytics.swift
+++ b/Sources/PrimerSDK/Classes/Core/Analytics/Analytics.swift
@@ -32,10 +32,11 @@ class Analytics {
         var primerAccountId: String?
         var properties: AnalyticsEventProperties? = nil
         var sdkSessionId: String
-        var sdkType: String = PrimerSource.sdkSourceType.sourceType
+        var sdkType: String
         var sdkVersion = Bundle.primerFramework.releaseVersionNumber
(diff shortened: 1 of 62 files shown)


Sources/PrimerSDK/Classes/Core/Analytics/Analytics.swift
```diff
@@ -32,10 +32,11 @@ class Analytics {
         var primerAccountId: String?
         var properties: AnalyticsEventProperties? = nil
         var sdkSessionId: String
-        var sdkType: String = PrimerSource.sdkSourceType.sourceType
+        var sdkType: String
         var sdkVersion = Bundle.primerFramework.releaseVersionNumber
         var sdkIntegrationType: PrimerSDKIntegrationType? = PrimerInternal.shared.sdkIntegrationType
         var sdkPaymentHandling: PrimerPaymentHandling? = PrimerSettings.current.paymentHandling
+        var integrationType: String
```
xevious78, line 39: We should try to avoid modifying the schema of the main analytics object 🤔 
Are you sure this is surfaced on metabase?

EvansPie: It's in the body of the event in Metabase, it's not mapped though at the moment (along with other vars). The above has been discussed and agreed, and it's already there in the released RC.

xevious78: thanks!

xevious78 (approved):

PR #592, opened 2023-03-22
Hotfix - Revisit analytics

EvansPie:


Sources/PrimerSDK/Classes/Core/Analytics/AnalyticsEvent.swift
```diff
@@ -66,6 +66,7 @@ extension Analytics.Event {
         }
         
         enum MessageType: String, Codable {
+            case analytics                          = "ANALYTICS"
```
xevious78, line 69: what is this used for?

xevious78 (approved):
one question

Sources/PrimerSDK/Classes/Core/Analytics/AnalyticsEvent.swift
EvansPie: Nowhere at the moment, I removed it. It was used while debugging to send an event with how many events were synced, and how many were still left locally.

PR #594, opened 2023-03-24
CHKT-1175 throw warning when decision handler is not called

EvansPie:
- Throw a console warning if `decisionHanlder` is not called within 5 sec


semirp:
[question] Would it be better/easier to `stop` the timer in case there's a call to `decisionHandler` instead of using a `decisionHandlerHasBeenCalled` flag?

PR #630, opened 2023-06-20
CHKT-1394 Silence "low-hanging fruit" warnings

borisprimer:


Internal/Debug App/Sources/Network/Networking.swift
```diff
@@ -96,11 +96,10 @@ class Networking {
                         
         msg += "Headers:\n\(request.allHTTPHeaderFields ?? [:])\n"
                 
-        if let body = body {
+        if let body = body,
+           let bodyJson = try? JSONSerialization.jsonObject(with: body, options: .allowFragments){
             request.httpBody = body
-            
-            let bodyJson = try? JSONSerialization.jsonObject(with: body, options: .allowFragments)
-            msg += "Body:\n\(bodyJson ?? [:])\n"
```
EvansPie, line 103: This was written this way as to always have console log and know what's happening, even if the body was `nil`.

Internal/Debug App/Sources/View Controllers/MerchantResultViewController.swift
```diff
@@ -42,11 +42,7 @@ class MerchantResultViewController: UIViewController {
         if logs.count > 0 {
             if let data = try? JSONSerialization.data(withJSONObject: logs) {
                 if let prettyNSStr = data.prettyPrintedJSONString {
-                    if let prettyStr = prettyNSStr as? String {
```
EvansPie, line 45: This is needed in the case of some failures of WebDriverIO (used in UI tests).

Internal/Debug App/Sources/View Controllers/MerchantResultViewController.swift
```diff
@@ -57,14 +53,11 @@ class MerchantResultViewController: UIViewController {
...
         }
 
+
         if let checkoutData {
             if let data = try? JSONEncoder().encode(checkoutData) {
                 if let prettyNSStr = data.prettyPrintedJSONString {
-                    if let prettyStr = prettyNSStr as? String {
-                        responseTextView.text = prettyStr
-                    } else {
-                        responseTextView.text = "[\"Failed to case NSString to String [WebDriverIO]\"]"
```
EvansPie, line 66: Same as above

Internal/Debug App/Sources/Network/Networking.swift
```diff
@@ -262,7 +261,6 @@ class Networking {
         let url = environment.baseUrl.appendingPathComponent("/api/client-session")
 
         var bodyData: Data!
-        var headers: [String: String]?
```
EvansPie, line 265: The reason we had those here is for when we want to test something in particular and need to pass some extra headers.

EvansPie (changes requested):

Internal/Debug App/Sources/Network/Networking.swift
borisprimer: OK, reversed

Internal/Debug App/Sources/View Controllers/MerchantResultViewController.swift
borisprimer: reversed.

Internal/Debug App/Sources/View Controllers/MerchantResultViewController.swift
borisprimer: reversed.

EvansPie (approved):

PR #631, opened 2023-06-21
CHKT-1374 Add few tests for Headles Vault Manager

borisprimer:
Write unit tests on headless vault manager https://primerapi.atlassian.net/browse/CHKT-1374


EvansPie:
@borisprimer can you also add a test on deleting a vaulted payment method?

PR #634, opened 2023-07-04
CHKT-1415 Vipps fix

EvansPie:
Fix Vipps getting cancelled


Sources/PrimerSDK/Classes/User Interface/TokenizationViewModels/WebRedirectPaymentMethodTokenizationViewModel.swift
```diff
@@ -316,6 +316,8 @@ class WebRedirectPaymentMethodTokenizationViewModel: PaymentMethodTokenizationVi
 extension WebRedirectPaymentMethodTokenizationViewModel: SFSafariViewControllerDelegate {
     
     func safariViewControllerDidFinish(_ controller: SFSafariViewController) {
+        if UIApplication.shared.applicationState != .active { return }
```
xevious78, line 319: Could you add a comment explaining why this is so important?
Wanna make sure we don't break anything by removing it in the future.

PR #657, opened 2023-08-08
Upgrade 3DS to the latest version

borisprimer:


Internal/Debug App/Podfile
```diff
@@ -1,13 +1,13 @@
 # Uncomment the next line to define a global platform for your project
-platform :ios, '10.0'
+platform :ios, '11.0'
```
semirp, line 2: Ideally we should keep the same version as before and bump minimum supported version on the Primer3DS side.

Also we should check the version when initializing 3DS service and throw an error in case merchants are using `2.0.0` of Primer3DS and user is on iOS 10.

borisprimer: Screenshot 2023-08-09 at 11 08 53

You can't run pod install because of the minimum iOS version in Podfile.

borisprimer: here is the other part of this task https://github.com/primer-io/primer-sdk-3ds-ios/pull/13

NQuinn27:
This can probably be closed now @BorisNikolic? I think we are up to date now on master

borisprimer:
> This can probably be closed now @BorisNikolic? I think we are up to date now on master

To run the 3DS SDK 2.0.0 minimum iOS version is 13.1. This is just a change on the example project Podfile to do that.
Should we go for minimum iOS support bump on Primer SDK as well as we mentioned before?

PR #659, opened 2023-08-24
Run unit tests during pipeline build

NQuinn27:
This PR adds running of the Debug App's Unit Test target before the Appetize build step.

Changes made to the unit tests:
- Moved an incorrectly placed expectation fulfill in `HeadlessVaultManager.test_headless_delete_vaulted_payment_method`
- Flipped an assertion (see comment in file)
- Updated expectation timeouts where the value was too low for the slower build machine

### DIFF
--- a/Internal/Debug App/Tests/Unit Tests/Primer/PrimerBancontactCardDataManagerTests.swift
+++ b/Internal/Debug App/Tests/Unit Tests/Primer/PrimerBancontactCardDataManagerTests.swift
@@ -13,6 +13,8 @@ import XCTest
 
 class PrimerBancontactCardDataManagerTests: XCTestCase {
     
+    private static let expectationTimeout = 1.0
+    
     func test_valid_raw_bancontact_card_data() throws {
         let exp = expectation(description: "Await validation")
         
@@ -34,7 +36,7 @@ class PrimerBancontactCardDataManagerTests: XCTestCase {
             exp.fulfill()
         }
         
-        wait(for: [exp], timeout: 0.1)
+        wait(for: [exp], timeout: expectationTimeout)
     }
     
     // We are making the below tests as well to make sure that the standards validation of simple card data passes
@@ -60,7 +62,7 @@ class PrimerBancontactCardDataManagerTests: XCTestCase {
             exp.fulfill()
         }
         
-        wait(for: [exp], timeout: 0.1)
+        wait(for: [exp], timeout: expectationTimeout)
     }
     
     func test_invalid_cardnumber_in_raw_card_data() throws {
@@ -85,7 +87,7 @@ class PrimerBancontactCardDataManagerTests: XCTestCase {
             exp.fulfill()
         }
         
-        wait(for: [exp], timeout: 0.1)
+        wait(for: [exp], timeout: expectationTimeout)
         exp = expectation(description: "Await validation")
         
         firstly { () -> Promise<Void> in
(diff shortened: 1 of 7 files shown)


Internal/Debug App/Tests/Unit Tests/v2/HeadlessVaultManagerTests.swift
```diff
@@ -335,7 +335,8 @@ final class HeadlessVaultManagerTests: XCTestCase {
                                         XCTAssert(true, "Failed with error \(err.localizedDescription)")
                                         
                                     } else if let updatedVaultedPaymentMethods = updatedVaultedPaymentMethods {
-                                        XCTAssert(!updatedVaultedPaymentMethods.isEmpty, "Failed to delete vaulted payment method")
+                                        XCTAssert(updatedVaultedPaymentMethods.isEmpty, "Failed to delete vaulted payment method")
+                                        exp.fulfill()
                                     } else {
```
NQuinn27, line 340: 1: This assertion I believe needed to be flipped- the test now passes.
2: With the expectation outside of these callbacks as it was, we never got this far to test the assertion.

@BorisNikolic please confirm my assumptions 🙏🏻

jnewc:
@NQuinn27 we had a quick chat about this already but will put here for visibility: as we're not sure if/when we'll decouple the tests from CocoaPods / the example app as you described, I think we should leave it as is for now and revisit when we've a clearer on project structure.

Internal/Debug App/Tests/Unit Tests/Primer/PrimerBancontactCardDataManagerTests.swift
```diff
@@ -13,6 +13,8 @@ import XCTest
 
 class PrimerBancontactCardDataManagerTests: XCTestCase {
     
+    private static let expectationTimeout = 1.0
```
jnewc, line 16: Worth making global? Thinking we can override for individual suites or tests later if they need it and a global makes it easier to change in one place when an Xcode release causes simulator-wide overheads (might be overthinking this :-D)

Internal/Debug App/Tests/Unit Tests/v2/HeadlessVaultManagerTests.swift
jnewc: `updatedVaultedPaymentMethods` should return the mocked result per line 331 whcih is an empty array, so yes this is right afaict.

fastlane/Fastfile
```diff
@@ -99,6 +102,9 @@ platform :ios do
       podfile: "Internal/Debug App/Appetize Podfile/Podfile"
     )
 
+    # Run Unit Tests
+    tests
```
jnewc, line 106: Does the `tests` step not build anyway? 

If so we could remove the `build_app` step below

NQuinn27: Good shout, let me check. If the `run_tests` lane can take all of the params `build_app` does, we can maybe reduce this to one step.

Internal/Debug App/Tests/Unit Tests/Primer/PrimerBancontactCardDataManagerTests.swift
NQuinn27: It did cross my mind as it is the same in two places right now, but I held off as there are expectation timeouts of various lengths throughout the tests, all depends on the work going on in the test.

Feels like a per-file value to me

jnewc: Yikes, agreed - another one to note for discussion (600s !?)

jnewc (approved):

borisprimer (approved):

xevious78 (approved):

PR #660, opened 2023-08-31
Update 3DS Version in Debug and Appetize Podfiles

NQuinn27:
DEX-

# What this PR does

Updates the 3DS Version in both the Debug app and Appetize Apps podfiles

# Before merging

_QA_

_Documentation_

# After merging

### DIFF
--- a/.github/workflows/build_test_upload.yml
+++ b/.github/workflows/build_test_upload.yml
@@ -74,7 +74,7 @@ jobs:
       - name: Create the Xcode project and workspace
         run: sh ./Internal/Debug\ App/tuist-generate.sh is_ci
 
-      - name: Distribute app to Appetize 🚀
+      - name: Test, Build, and Distribute app to Appetize 🚀
         run: |
           bundle exec fastlane appetize_build_and_upload
         env:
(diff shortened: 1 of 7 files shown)


jnewc (approved):
LGTM - but wondering why we have a separate + almost identical Podfile for Appetize

NQuinn27:
@jnewc it looks to me that it is to do with the ipay88 pod failing on arm64. Perhaps appetize doesnt use the rosetta simulator? 
Either way, I think we can remove this soon - looks like the iPay88 library has been updated so we can build without rosetta.

Internal/Debug App/Appetize Podfile/Podfile
```diff
@@ -1,5 +1,5 @@
 # Uncomment the next line to define a global platform for your project
-platform :ios, '10.0'
+platform :ios, '13.0'
```
borisprimer, line 2: The minimum should be 13.1 as stated in .podspec of the ThreeDS lib we are using.

borisprimer (changes requested):

Internal/Debug App/Appetize Podfile/Podfile
NQuinn27: Updated, thanks!

borisprimer (approved):

PR #661, opened 2023-09-01
Remove PollingURLs from PollingResponse

jnewc:
CHKT-1636

Removes `PollingURLs` from `PollingResponse` per conversation with @xevious78.

# Before merging

_QA_

- [x] I manually tested this with a demo application (simulator

# After merging


borisprimer (approved):

NQuinn27:
@xevious78  Are merchants are accessing `PollingResponse.urls` directly in their own code right now?
Im wondering if we should remove it entirely or leave it as an empty object with a deprecation notice for a period. Something like `@available(*, deprecated, message: "This property will be removed in the future.")`

I would be in favour of ripping the bandaid now to avoid creating future work for ourselves, but if it is expected to create a problem for a lot of merchants, the gradual approach might be preferable.

NQuinn27 (approved):

PR #665, opened 2023-09-06
Remove hard-coded version check for SDK version analytics event

jnewc:
DEX-

# What this PR does

Previously we've been performing a hard-coded check on the SDK version and reporting it via analytics. The purpose of this was to investigate an issue caused by the wrong version being reported.

I've made two changes:
1. Consolidated the two investigation events into a single version report event
2. Switched to using `bundle(for: <type>)` to ensure that we are always getting the version from the SDK module

The event can be viewed in the Xcode debug console while running the Debug app

# Before merging

_QA_


NQuinn27:
Were you able to reproduce at all @jnewc? I'm wondering if there is something we are missing regarding the issues with versions being reported wrong in the wild.

- When investigating [this issue](https://primerapi.slack.com/archives/C01RYAZQ6D6/p1693987082774689) for @semirp,  I cant reproduce, the versions are correct for me both with `Bundle(for:)` and `Bundle(identifier:)`.
- @BorisNikolic do you have some context on this [change](https://github.com/primer-io/primer-sdk-3ds-ios/commit/f1e207f30f5f8db652c126115a4dac393f6ad03d) (goes the opposite direction to this one).

jnewc:
@NQuinn27 I couldn't reproduce but this approach won't work for SPM anyway, so we definitely need to rethink it.

My experience of using `Bundle(for:)` in the past is that it will always return the identifier for the package in context so we should also use that for CocoaPods integrations.

In short, there's more to do here but this is a step in the right direction.

xevious78 (approved):

Sources/PrimerSDK/Classes/Core/Primer/PrimerInternal.swift
```diff
@@ -101,25 +101,18 @@ internal class PrimerInternal {
...
-                eventType: .message,
-                properties: MessageEventProperties(
-                    message: "Wrong release version number (\(bundleReleaseVersionNumber ?? "n/a")) detected.",
-                    messageType: .error,
-                    severity: .error))
-            )
-        } else {
-            events.append(Analytics.Event(
+        
+        let bundleReleaseVersionNumber = Primer.bundle.releaseVersionNumber
```
NQuinn27, line 105: If you look inside `BundleExtension.swift` you will see that using this reverts to the cocoapods bundle ID, via `Bundle.primerFramework`. Think we need to update the code there too.

Poorly named instance method behaving as a static method...

NQuinn27 (commented):
Agree on the bundle change it is a step in the right direction. Looks good to me, just one comment.

Sources/PrimerSDK/Classes/Core/Primer/PrimerInternal.swift
jnewc: Oof, good point. I'm not against having it as a static property ultimately (similar to `Bundle.main`) but we should have a single point of entry. 

With that in mind, I'll remove `Primer.bundle` and update `Bundle.primerFramework`, and point the analytics code at that 👍

NQuinn27 (approved):

borisprimer (approved):

PR #669, opened 2023-09-12
chore: updated Debug app for latest UI Tests

NQuinn27:
DEX-

# What this PR does
Various changes to the Debug App to facilitate UI tests
- Adds "Disable init screen" toggle to settings screen
- Adds Session Intent chooser to Drop-in and Headless UI
- Adds Payment Method Type manual input flow to Drop-in UI
- Adds Dev environment to the Environment Chooser
- Adds mock 3DS service

Browserstack tests are run off the companion branch in mobile-appium-tests until those are merged into master.

The mock 3DS implementation was half-complete in the Debug app, it must have been broken in the `2.17.0` merge. I chose to move the mocking one level up (to the `CardFormPaymentMethodTokenizationViewModel`). This removes all test code from `3DSService` and replaces it with a mock service to show the required test UI. Ping me if you would like a client session token to see that in action.

Run appium tests locally, or view the actions logs
Please test the real 3DS flow

# Before merging

_QA_


NQuinn27:
Ready for review but expect one more change when [this PR](https://gitlab.com/primer-io/dx/mobile-appium-tests/-/merge_requests/5) is merged, I will update the build script to point to develop on the appium tests

Debug App/Sources/View Controllers/MerchantDropInUIViewController.swift
```diff
@@ -81,72 +88,52 @@ class MerchantDropInUIViewController: UIViewController, PrimerDelegate {
...
-                    let merchantErr = NSError(domain: "merchant-domain", code: 1, userInfo: [NSLocalizedDescriptionKey: "Failed to fetch client token"])
-                    print(merchantErr)
-                } else if let clientToken = clientToken {
-                    Primer.shared.showPaymentMethod("KLARNA", intent: .vault, clientToken: clientToken)
-                }
-            }
-        } else {
-            fatalError()
+    
+    @IBAction func paymentMethodSessionIntentChanged(_ sender: UISegmentedControl) {
```
jnewc, line 119: We haven't included the `KLARNA` method from above in this - was it just not needed or is to be reintroduced?

jnewc (approved):

borisprimer (approved):

xevious78 (approved):
Approving it for now, but keen to define the standard here.
- Are we all aligned to "pollute" the code with `#if DEBUG` statement and `testId`? Is there a way to inject this code from the compiler outside?
- What if somebody passes this flag at build time?

NQuinn27:
> Approving it for now, but keen to define the standard here.
> 
> * Are we all aligned to "pollute" the code with `#if DEBUG` statement and `testId`? Is there a way to inject this code from the compiler outside?
> * What if somebody passes this flag at build time?

Thanks @xevious78. 

We have been discussing this in the iOS Chapter. It is not yet the ideal solution, we do not want to continue polluting the code with `#DEBUG` statements, we would preferably inject this from the top level, but it will be a process which begins with agreeing on the Dependency Injection system we will use.

This PR does bring the test code up one level and removes it entirely from the `3DSService` and replaces it with a simple mock, which is a step in the right direction.

In terms of safety, the #DEBUG flag can be set by a merchant, but without the `testId` in the Client Session response, the test code will not be executed.

We are on the same page - we want to continue to make this cleaner.

PR #681, opened 2023-09-26
chore: use version.swift file

NQuinn27:
# What this PR does
- Replaces all usage of `info.plist` for version reporting with the new `version.swift`
- Added a utility to wrap this number to preserve current react native version logic
- Added unit tests to this utility

# Before merging

_QA_

### DIFF
--- a/Sources/PrimerSDK/Classes/Extensions & Utilities/VersionUtils.swift
+++ b/Sources/PrimerSDK/Classes/Extensions & Utilities/VersionUtils.swift
@@ -1,0 +1,8 @@
+//
+//  VersionUtils.swift
+//  PrimerSDK
+//
+//  Created by Niall Quinn on 26/09/23.
+//
+
+import Foundation
--- a/Sources/PrimerSDK/Classes/Extensions & Utilities/VersionUtils.swift
+++ b/Sources/PrimerSDK/Classes/Extensions & Utilities/VersionUtils.swift
@@ -6,3 +6,20 @@
 //
 
 import Foundation
+
+struct VersionUtils {
+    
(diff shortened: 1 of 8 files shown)


borisprimer (approved):

Sources/PrimerSDK/Classes/Extensions & Utilities/VersionUtils.swift
```diff
@@ -0,0 +1,25 @@
...
+struct VersionUtils {
+    
+    /**
+     Returns the version string in the format `"major.minor.patch"`
+     
+     If `PrimerIntegrationOptions.reactNativeVersion` is set, it will be returned.
+     If not, the version specified as `PrimerSDKVersion` in the file `"sources/version.swift"` will be returned.
+     */
+    static var releaseVersionNumber: String? {
+        if let reactNativeVersion = Primer.shared.integrationOptions?.reactNativeVersion {
```
borisprimer, line 19: Just a small question, sorry if I missed this in Slack chat, but where do we get the `reactNativeVersion` string from?

NQuinn27: Its provided by the React Native wrapper via `integrationOptions`, [here](https://github.com/primer-io/primer-sdk-react-native/blob/c321b86c3dc71597c2d6c798202dfc20bff326ab/ios/Sources/RNTPrimer.swift#L77).

NQuinn27:
> Would be great to have a similar mechanism for the other SDKs

100% - the wrappers will be getting this treatment along with the commitizen release flow work

PR #685, opened 2023-09-28
fix: fixes ApplePay button rendering logic

borisprimer:
Fixed a bug and added a few more config options for SDK, to show/hide ApplePay if the device is not capable of doing it, etc.

[RFC](https://www.notion.so/primerio/RFC-for-Enhancing-ApplePay-Flow-in-PrimerSDK-iOS-framework-7afe70b4a2c248a7819a925c90210573)


Debug App/Sources/View Controllers/MerchantSessionAndSettingsViewController.swift
```diff
@@ -224,6 +224,7 @@ class MerchantSessionAndSettingsViewController: UIViewController {
             environmentSegmentedControl.selectedSegmentIndex = 1
         }
         
+        customDefinedApiKey = "ccee2be8-df08-4c79-856d-40085349e9da"
         self.apiKeyTextField.text = customDefinedApiKey
```
NQuinn27, line 228: These three changes:
- Changing default env to .staging
- Changing default session currency to .AED
- Hardcoding an apiKey

Are they intended?

Sources/PrimerSDK/Classes/Core/PrimerHeadlessUniversalCheckout/Models/PrimerHeadlessUniversalCheckoutPaymentMethod.swift
```diff
@@ -8,15 +8,26 @@
...
     
     public class PaymentMethod: NSObject {
         
         static var availablePaymentMethods: [PrimerHeadlessUniversalCheckout.PaymentMethod] {
-            let availablePaymentMethods = PrimerAPIConfiguration.paymentMethodConfigs?
+            var availablePaymentMethods = PrimerAPIConfiguration.paymentMethodConfigs?
                 .compactMap({ $0.type })
                 .compactMap({ PrimerHeadlessUniversalCheckout.PaymentMethod(paymentMethodType: $0) })
+            
+            if !PrimerSettings.current.paymentMethodOptions.applePayOptions!.showApplePayForUnsupportedDevice {
```
NQuinn27, line 22: Are we confident that this force unwrap is safe? It might be better to use ? here

Sources/PrimerSDK/Classes/Data Models/PrimerConfiguration.swift
```diff
@@ -101,7 +101,14 @@ extension Response.Body {
             ?? []
             
             let supportedNetworks = PaymentNetwork.iOSSupportedPKPaymentNetworks
-            if !PKPaymentAuthorizationViewController.canMakePayments(usingNetworks: supportedNetworks) {
+            var canMakePayment: Bool
+            if PrimerSettings.current.paymentMethodOptions.applePayOptions!.checkProvidedNetworks {
```
NQuinn27, line 105: Same goes for this force unwrap, is it safe?

NQuinn27 (commented):
Nice work, a few comments

Debug App/Sources/View Controllers/MerchantSessionAndSettingsViewController.swift
borisprimer: oooops!

jnewc (commented):
Is there a reason it isn't filtered out as a payment method as the source? i.e. in `listAvailablePaymentMethodsTypes`

I think this doesn't fix the problem we discussed with different parts of the flow having a different outcome when querying whether Apple Pay is available (see diagram)

image

As I mentioned previously, `listAvailablePaymentMethodsTypes` has exactly the same filtering logic as `paymentMethodConfigViewModels` _except_ for the ApplePay filtering. I don't think we want to leave this inconsistency in place.

Debug App/Sources/View Controllers/MerchantSessionAndSettingsViewController.swift
```diff
@@ -447,7 +447,9 @@ class MerchantSessionAndSettingsViewController: UIViewController {
                 applePayOptions: PrimerApplePayOptions(
                     merchantIdentifier: "merchant.checkout.team",
                     merchantName: merchantNameTextField.text ?? "Primer Merchant",
-                    isCaptureBillingAddressEnabled: false)),
+                    isCaptureBillingAddressEnabled: false,
+                    showApplePayForUnsupportedDevice: false,
+                    checkProvidedNetworks: false)),
```
jnewc, line 452: I think this needs to specify it's for apple pay, e.g. `checkProvidedNetworksAvailableForApplePay` (or something like that)

jnewc (approved):

NQuinn27 (approved):

PR #694, opened 2023-09-29
feat: Logging interface for SDK

jnewc:
# What this PR does

* Introduce a logging interface that can be implemented by merchants
* Implements it within the debug app

_QA_

_Documentation_

# After merging


Debug App/Sources/Network/Networking.swift
```diff
@@ -100,27 +100,27 @@ class Networking {
...
-        print(msg)
+        Primer.shared.logger?.debug(message: msg)
         msg = ""
         
         URLSession.shared.dataTask(with: request, completionHandler: { (data, response, err) in
             DispatchQueue.main.async {
-                msg += "RESPONSE\n"
-                msg += "URL: \(request.url?.absoluteString ?? "Invalid")\n"
+//                msg += "RESPONSE\n"
+//                msg += "URL: \(request.url?.absoluteString ?? "Invalid")\n"
```
NQuinn27, line 112: These items which are commented out: leftover from development, or do we want to remove them entirely>

Debug App/Tests/Unit Tests/Analytics/AnalyticsTests.swift
```diff
@@ -1445,14 +1445,7 @@ class AnalyticsTests: XCTestCase {
...
         } catch {
-            primerLogAnalytics(
-                title: "ANALYTICS",
-                message: error.localizedDescription,
-                prefix: "📚",
-                bundle: Bundle.primerFrameworkIdentifier,
-                file: #file, className: "\(Self.self)",
-                function: #function,
-                line: #line)
+            print("📚 ANALYTICS\n📚 \(error.localizedDescription)")
```
NQuinn27, line 1448: Is this a special case because it exists only in unit tests? Is it critical for all the tests that this file is created correctly?
I wonder if we should just fail the test with a nice description here...

Definitely a nitpick - aware this requires some refactoring to make this method throw - happy to let it go in this case.

Debug App/Sources/Network/Networking.swift
jnewc: These are to be removed 👍 Generally, I've tried to avoid multiline logs as they don't present well in the console + we should try to keep logs discrete so they are more easily filterable in the future

Debug App/Sources/View Controllers/MerchantHeadlessCheckoutCardComponentsViewController.swift
```diff
@@ -60,7 +60,7 @@ class MerchantHeadlessCheckoutCardComponentsViewController: UIViewController {
         self.stackView.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor, constant: 20).isActive = true
         self.stackView.trailingAnchor.constraint(equalTo: view.trailingAnchor, constant: -10).isActive = true
                 
-        self.cardComponentsManager = try! PrimerHeadlessUniversalCheckout.CardComponentsManager()
+        self.cardComponentsManager = PrimerHeadlessUniversalCheckout.CardComponentsManager()
```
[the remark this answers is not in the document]
jnewc: Yeah, I was surprised it allows redundant `try`s. It does throw a warning but it was somewhat hidden by the deprecation warning.

borisprimer (approved):
LGTM! Great work!

NQuinn27 (approved):

Sources/PrimerSDK/Classes/Core/Logging/PrimerLogger.swift
```diff
@@ -0,0 +1,114 @@
...
+        switch self {
+        case .debug: return "🪲"
+        case .info: return "ℹ️"
+        case .warning: return "⚠️"
+        case .error: return "🚨"
+        }
+    }
+}
+
+public protocol PrimerLogger {
```
jnewc, line 31:
```suggestion
public protocol PrimerLogger {
    var logLevel: LogLevel { get }
```

Debug App/Sources/Utility/ExampleLogger.swift
```diff
@@ -0,0 +1,51 @@
...
+//  Created by Jack Newcombe on 13/09/2023.
+//  Copyright © 2023 Primer API Ltd. All rights reserved.
+//
+
+import Foundation
+import PrimerSDK
+import OSLog
+
+class ExampleLogger: PrimerLogger {
+
```
jnewc, line 14:
```suggestion
    
    var logLevel: LogLevel = .none
```

Sources/PrimerSDK/Classes/Core/Logging/PrimerLogger.swift
```diff
@@ -0,0 +1,175 @@
...
+    /// Indicates which logs should be received by the logger.
+    ///
+    /// For example:
+    ///  - setting a level of `info` will result in all `info`, `warning` and `error` logs being received, but no debug logs.
+    ///  - setting a level of `none` will result in no logs being received
+    var logLevel: LogLevel { get set }
```
semirp, line 42: 👍

Sources/PrimerSDK/Classes/Core/3DS/3DSService.swift
```diff
@@ -413,7 +413,7 @@ class ThreeDSService: ThreeDSServiceProtocol {
                     // All good, url value is valid and https
                     threeDsAppRequestorUrl = url
                 } else {
-                    print("PRIMER WARNING!\nthreeDsAppRequestorUrl is not in a valid format (\"https://applink\"). In case you want to support redirecting back during the OOB flows please set correct threeDsAppRequestorUrl in PrimerThreeDsOptions during SDK initialization.")
+                    logger.warn(message: "Warning: threeDsAppRequestorUrl is not in a valid format (\"https://applink\"). In case you want to support redirecting back during the OOB flows please set correct threeDsAppRequestorUrl in PrimerThreeDsOptions during SDK initialization.")
```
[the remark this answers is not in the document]
jnewc: We have a prefix (⚠️, 🚨, etc.) that is added to logs for both OSLog and `print`, and it also highlights logs in the relevant background colour when using OSLog. Let me grab some screenshots for clarity.

jnewc:
>General question, if developer only use headless integration, is it enough to set Logger only on Primer object?

Yes, the logger works independently of checkout type. This is why the `logger` property is on the SDK singleton (`Primer.shared.logger`)

Sources/PrimerSDK/Classes/Core/3DS/3DSService.swift
semirp, line 416:
```suggestion
                    logger.warn(message: "threeDsAppRequestorUrl is not in a valid format (\"https://applink\"). In case you want to support redirecting back during the OOB flows please set correct threeDsAppRequestorUrl in PrimerThreeDsOptions during SDK initialization.")
```

sorry, I meant this :)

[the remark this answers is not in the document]
jnewc: OSLogger:

image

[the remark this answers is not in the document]
jnewc: `print`:

image

Sources/PrimerSDK/Classes/Core/Logging/LogReporter.swift
```diff
@@ -0,0 +1,24 @@
...
+protocol LogReporter {}
+extension LogReporter {
+    
+    var logger: PrimerLogger {
+        return PrimerLogging.shared.logger
+    }
+    
+    static var logger: PrimerLogger {
+        return PrimerLogging.shared.logger
+    }
```
borisprimer, line 21: What's the difference between the two?

borisprimer (approved):

Sources/PrimerSDK/Classes/Core/Logging/LogReporter.swift
jnewc: The latter is implicitly available to static functions, so instead of calling e.g. `PrimerLogging.shared.logger.info` in a static function body, you can call `logger.info`.

Plus, you're guaranteed the same concrete log reporter in both static and non-static functions

jnewc: Example here: https://github.com/primer-io/primer-sdk-ios/pull/694/commits/d9e055ab9dbf7442ba8683291791c7de4aca7fe6#diff-8e672db59ad5703e14562620d8d5507628aa14e8d0154c166d23ce65e3898565L33

NQuinn27 (approved):

PR #698, opened 2023-10-05
Release 2.17.6

primer-security-integrations:
Bumped 2.17.5 -> 2.17.6


borisprimer (approved):

CHANGELOG.md
semirp, line 1: Nice!

.gitlab-ci.yml
```diff
@@ -0,0 +1,15 @@
+variables:
+  RUBY_VERSION: 3.2.2
+
+# workflow:
+#   rules:
+#     - if: $CI_COMMIT_TAG
+
+build-and-release:
```
semirp, line 8: Should we add check to this job so it triggers only when there's a new tag?
`rules:`
    `- if: $CI_COMMIT_TAG`

NQuinn27: Yes, the trigger should be as you say "when there is a new tag".
Right now it is manual, so we will trigger this release with a button in Gitlab... 

I thought its better to have some more control until we know it works, what do you think?

NQuinn27: My misunderstanding above. It will be manual and only be available when there is a new tag pushed with your proposed changes.
Thanks for clearing that up async @semirp - changes made

semirp (approved):

PR #702, opened 2023-10-16
chore: Update iOS deployment target to 14.0 in debug app project + podfile

jnewc:
CHKT-1763

# Before merging

_QA_


NQuinn27 (approved):
LGTM

jnewc:
Passed [here](https://github.com/primer-io/primer-sdk-ios/actions/runs/6533648849/job/17739468938), and there's only one commit ... will re-run to see what's happening but merging after that if it's unrelated flakiness

NQuinn27:
> Passed [here](https://github.com/primer-io/primer-sdk-ios/actions/runs/6533648849/job/17739468938), and there's only one commit ... will re-run to see what's happening but merging after that if it's unrelated flakiness

My most recent run [also failed](https://github.com/primer-io/primer-sdk-ios/actions/runs/6534200086?pr=703), so I think we had some flakiness. Perhaps even caused by concurrent runs? (Which we will address when moving to nightly UI tests this sprint)

PR #715, opened 2023-10-23
chore: Report min deployment version as semver string with every event

jnewc:
CHKT-1683

# What this PR does

Reports the minimum deployment target (i.e. min required OS version) of the integrating app.

Inspect the event at runtime

# Before merging

_QA_

### DIFF
--- a/Debug App/Tests/Unit Tests/Analytics/AnalyticsTests.swift
+++ b/Debug App/Tests/Unit Tests/Analytics/AnalyticsTests.swift
@@ -1491,31 +1491,6 @@ class AnalyticsTests: XCTestCase {
             Analytics.Service.deleteAnalyticsFile()
         }
     }
-    
-    func testMinDeploymentTargetParsesCorrectly() {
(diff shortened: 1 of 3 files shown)


NQuinn27:
Forgive me if I am missing something - but I tested this using the Debug App, and it appears to report the min deployment version of the SDK, not the app itself - i.e, I see `po self.minDeploymentTarget "10.0.0"`. I would expect `13.1` in the Debug App

jnewc:
@NQuinn27 you're correct. Unfortunately it looks like CocoaPods produces a different answer to the SPM project that I tested against (which returns the expected `12.0.0`).

I think the alternative route is to query the info.plist of the integrating project. I'll give it a try and see if it's more consistent.

jnewc:
@NQuinn27 using `MinimumOSVersion` from the main bundle's info.plist works for both CocoaPods and SPM ✅

NQuinn27:
> @NQuinn27 using `MinimumOSVersion` from the main bundle's info.plist works for both CocoaPods and SPM ✅

Nice, tested and looks good

NQuinn27 (approved):

PR #721, opened 2023-10-31
chore: builds app on multiple versions of xcode nightly

NQuinn27:
- Adds a workflow to build the app using multiple versions of xcode, nightly
- Chose 14.3.1, 14.2, 14.1 as it covers all Xcode versions which can submit builds to the App Store
- Refactors Fastfile to reuse cocoapods build code
- Reports failure on a per-build basis
- Reports success only if all builds succeed, to avoid spamming Slack

Successful test run: https://github.com/primer-io/primer-sdk-ios/actions/runs/6736276576

Example success report:
Screenshot 2023-11-03 at 11 06 43

Example failure report:
Screenshot 2023-11-02 at 18 55 46


.github/workflows/xcode-versions-build.yml
```diff
@@ -0,0 +1,74 @@
...
+
+jobs:
+  build:
+    name: "Build App"
+    runs-on: macos-13
+    strategy:
+      max-parallel: 3
+      matrix:
+        variant: [cocoapods, spm]
+        xcode: [14.3, 14.2, 14.1]
```
borisprimer, line 20: Why do we do it for 14.2 and 14.1? I would say 14.3.1 should be enough...

NQuinn27: My reasoning was covering all Xcode versions which can submit a build to the AppStore
Do you think it is safe to say if it builds on the latest Xcode 14 we are OK?

.github/workflows/ui-tests.yml
```diff
@@ -3,17 +3,17 @@ name: UI Tests
...
     types:
       - opened
       - edited
       - synchronize
-    branches:
-      - 'release/next'
 
 jobs:
   build-and-upload-to-firebase-and-browserstack:
+    if: startsWith(github.head_ref, 'release/') || github.head_ref == 'master'
```
NQuinn27, line 16: The trigger matches against the `base` not the head.
I believe this setup gets us:

Run on any PR pointed to master, only if the head is `release/**`
The check for `|| github.head_ref == 'master'` is to allow nightly runs to pass, and for manual runs (which must be triggered on master in this case)

Thoughts @HarryPrimer @semirp

semirp: `- run: echo "BRANCH_NAME=${{ github.event.pull_request && github.head_ref || github.ref_name }}" >> $GITHUB_ENV`

This is a good way to get the branch name in all cases, then it could work just to compare current branch new and what we want to execute.

I believe it could work if we only change `github.ref_name == 'master'`

NQuinn27: Good shout. But I see another problem here. It will run on any PR pointing to `master` currently, which is not what I want.

I want to satisfy all of these scenarios:
- Run on all pull requests with the head of `release/**`
- Allow nightly runs to run on `master`
- Allow manual runs to run on `master`

🤔

NQuinn27: Using `- run: echo "BRANCH_NAME=${{ github.event.pull_request && github.head_ref || github.ref_name }}"`, I believe we could use 

`if: $BRANCH_NAME == "release/next" || $BRANCH_NAME == "master"`

I think that is correct, do you think it is cleaner than the above? @semirp

semirp: Yeah that would work, sorry I wanted to say that it could work even if we just change it to:
` startsWith(github.head_ref, 'release/') || github.ref_name == 'master'`

but I believe targeting branches as you described could be even simpler

.github/workflows/xcode-versions-build.yml
jnewc: I think latest 14 is safe unless we start to build against specific deploy targets / swift versions too (which is food for thought!)

.github/workflows/xcode-versions-build.yml
```diff
@@ -0,0 +1,113 @@
...
+          bundler-cache: true
+      - name: Setup node
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18.3.0
+      - run: npm install --save slack-message-builder
+      - name: Build and archive app via Cocoapods
+        if: ${{ matrix.variant == 'cocoapods' }}
+        run: |
+          bundle exec fastlane build_cocoapods tests:false skip_package_ipa:false
```
jnewc, line 54: Is this for the purpose of making the builds faster, or some other reason?

Having it package would validate that the debug app works across Xcode versions too

jnewc (approved):

.github/workflows/xcode-versions-build.yml
NQuinn27: The opposite, this is ensuring we do package at the expense of build times. Note the variable name `skip_package_ipa`.

The Fastlane script previously used `skip_package_ipa=true` as a default to speed up builds. I allowed it to be set from the outside here so we can pick and choose

NQuinn27:
Updated this with a change to use the `push` trigger, which covers `release/**`
Verified in another PR: `release/testing_trigger` -> This one. https://github.com/primer-io/primer-sdk-ios/pull/726/checks

semirp (approved):

PR #722, opened 2023-11-02
chore: Calculate isDebug for Nol automatically

borisprimer:


PrimerSDK.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
     s.name         = "PrimerSDK"
-    s.version      = "2.18.0-b1"
+    s.version      = "2.18.0"
```
NQuinn27, line 3: We should leave this as it was and let commitizen do its thing

NQuinn27: Actually its fine to leave it, commitizen wont care what it is here

NQuinn27 (approved):

PR #723, opened 2023-11-02
chore: Add card validation logic for remote validation and co-badged cards

jnewc:
CHKT-1753

ℹ️ **This is going to a feature branch, not master**

# What this PR does

* Adds logic for calling the BIN data API during card entry
* Adds new delegate methods for async return of card networks during card entry
* Unit tests for the above

* Network layer has been modified to allow cancellation of requests

# Before merging

_QA_

- [x] I wrote unit tests


Sources/PrimerSDK/Classes/PCI/Checkout Components/PrimerCardData.swift
```diff
@@ -32,20 +32,32 @@ public class PrimerCardData: PrimerRawData {
...
+             cardNetworkIdentifier
     }
         
     public required init(
         cardNumber: String,
         expiryDate: String,
         cvv: String,
-        cardholderName: String?
+        cardholderName: String?,
+        cardNetworkIdentifier: String?
```
NQuinn27, line 54: Since `PrimerCardData` is public, it might be wise to give `cardNetworkIdentifier` a default value of nil, so as not to introduce a breaking change.

jnewc: Ah good catch, this was the intention but it slipped by me. Will change 👍

Sources/PrimerSDK/Classes/Core/PrimerHeadlessUniversalCheckout/Models/PrimerCardValidationModels.swift
```diff
@@ -0,0 +1,41 @@
...
+//  PrimerCardValidationModels.swift
+//  PrimerSDK
+//
+//  Created by Jack Newcombe on 30/10/2023.
+//
+
+import Foundation
+
+@objc
+public class PrimerCardValidationState: NSObject {
```
semirp, line 11: Should we consider calling this `PrimerCardState` instead of `PrimerCardValidationState`? As we are not validating  anything here

Sources/PrimerSDK/Classes/Core/BIN Data/CardValidationService.swift
```diff
@@ -0,0 +1,126 @@
...
+//  CardValidationService.swift
+//  PrimerSDK
+//
+//  Created by Jack Newcombe on 30/10/2023.
+//
+
+import Foundation
+
+protocol CardValidationService {
+    func validateCardNetworks(withCardNumber card5Number: String)
```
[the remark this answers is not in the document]
jnewc: No! Changing 😵‍💫

Sources/PrimerSDK/Classes/Core/PrimerHeadlessUniversalCheckout/Models/PrimerCardValidationModels.swift
jnewc: Yeah - I actually call the argument `cardState` in most places, so this makes sense. Will change 👍

Sources/PrimerSDK/Classes/Core/BIN Data/CardValidationService.swift
```diff
@@ -0,0 +1,126 @@
...
+            }
+            let cardMetadata = PrimerCardMetadata(availableCardNetworks: result.networks.map { network in
+                PrimerCardNetwork(displayName: network.displayName, networkIdentifier: network.value)
+            })
+            
+            self?.delegate?.primerRawDataManager?(rawDataManager,
+                                                  didReceiveCardMetadata: cardMetadata,
+                                                  forCardState: cardState)
+        }.catch { error in
+            // JN TODO: CHKT-1772 - send event
```
borisprimer, line 73: I usually write it like this:
// TODO: (BNI) comment

BNI is for (B)oris (NI)kolić of course

This way you have this feature enabled :) 
Screenshot 2023-11-07 at 08 02 04
 you save

Sources/PrimerSDK/Classes/Core/BIN Data/CardValidationService.swift
```diff
@@ -0,0 +1,126 @@
...
+    }
+    
+    // MARK: API Logic
+    
+    var validateCardNetworksCancellable: PrimerCancellable?
+    
+    private func listCardNetworks(_ cardNumber: String) -> Promise<Response.Body.Bin.Networks> {
+        
+        // ⚠️ We must only ever send eight or less digits to the endpoint
+        let cardNumber = String(cardNumber.prefix(8))
```
borisprimer, line 96: maybe extract this number 8 to some constants? don't bother doing it if you will never have repetition of this same number in same context

Sources/PrimerSDK/Classes/Core/PrimerHeadlessUniversalCheckout/Models/PrimerCardValidationModels.swift
```diff
@@ -0,0 +1,41 @@
...
+//  PrimerSDK
+//
+//  Created by Jack Newcombe on 30/10/2023.
+//
+
+import Foundation
+
+@objc
+public class PrimerCardState: NSObject {
+    public let cardNumber: String
```
borisprimer, line 12: If this object represents the current string from the card number text field (not complete, but while the user is typing), that has to be validated, maybe consider a different name. Here is what GPT suggests :) https://chat.openai.com/share/3ff353ef-09a3-46b0-b4ea-1a01aa097b82

borisprimer (approved):

Sources/PrimerSDK/Classes/Core/BIN Data/CardValidationService.swift
jnewc: Good shout - I'll do this as part of CHKT-1885 (Modify card entry logic to account for eight digit constraint)

Sources/PrimerSDK/Classes/Core/PrimerHeadlessUniversalCheckout/Models/PrimerCardValidationModels.swift
jnewc: Do you mean for the class name or the property name? I was previously using `cardFragment` for the property name but changed it to `cardNumber` as it clashed with the naming for existing local valdiation

@semirp what are you using in this case?

Sources/PrimerSDK/Classes/Core/BIN Data/CardValidationService.swift
jnewc: Ah nice, I have a saved search in Xcode to make sure I sort them all out before they get to master

Sources/PrimerSDK/Classes/Core/PrimerHeadlessUniversalCheckout/Models/PrimerCardValidationModels.swift
borisprimer: Class name... But it could be both for more context, class and property names could be different... Not a must, just a nice-to-have

jnewc: `CardNumberEntryState` looks good - I think this with a `Primer` suffix 👍

NQuinn27 (approved):

semirp (approved):

PR #726, opened 2023-11-03
Run on push to release/

NQuinn27:
Use `push` rather than pull request trigger


NQuinn27:
Im stupid @semirp - This works ☝️ 
Using `push` allows us to target a single branch, and it shows up in the PR as normal.

PR #727, opened 2023-11-06
chore: Ensure other linker flags are only set for included SDKs

jnewc:
CHKT-1849

# What this PR does

Ensures that we only append framework linker flags for extension SDKs that are present in the podfile

e.g. if `PrimerKlarnaSDK` is included, we append `-framework PrimerKlarnaSDK` to `OTHER_LDFLAGS`. Otherwise, it is omitted.

Install different combinations of Primer pods, and build

# Before merging

_QA_


borisprimer (approved):

Debug App/Podfile
```diff
@@ -45,10 +45,14 @@ end
...
   installer.generated_projects.each do |project|
+    primer_target_names = project.targets.filter { |target|
+      !target.name.start_with?("PrimerSDK") && target.name.start_with?("Primer")
+    }.map { |t| t.name }
     project.targets.each do |target|
       if target.name == "PrimerSDK"
         target.build_configurations.each do |config|
-          config.build_settings['OTHER_LDFLAGS'] = '$(inherited) -framework "Primer3DS" -framework "PrimerKlarnaSDK"'
+          other_ldflags = "$(inherited) -framework #{primer_target_names.join(' -framework ')}"
+          config.build_settings['OTHER_LDFLAGS'] = other_ldflags
```
NQuinn27, line 55: Previously the string was: `$(inherited) -framework "Primer3DS" -framework "PrimerKlarnaSDK"`
Now it will be `$(inherited) -framework Primer3DS -framework PrimerKlarnaSDK`

The linker has no problem with the lack of quotes?

jnewc: It depends. Each token is essentially a command line argument for `ld` so it only needs to be quoted if it contains spaces. By convention, pods don't include spaces in their name. We could add it for a bit of future proofing though - wdyt?

NQuinn27: Seems it doesnt, I wonder would it be better to be safe, in the map: `.map { |t| "\"#{t.name}\"" }`

jnewc:
Just noticed something: this will break if there are no other Primer* SDKs as it'll have a dangling `-framework` arg. Fixing.

NQuinn27 (approved):

PR #729, opened 2023-11-08
feat: Klarna Headless

ikhrypunov:
[NII-17](https://primer-ext.atlassian.net/browse/NII-17)

# What this PR does

The PR contains implementation of headless integration for Klarna payment system with changes to DebugApp related to integration of implemented solution

The PR contains integration of new version of PrimerKlarnaSDK with updated implementation, which uses new xcframework approach.

# Before merging

_QA_

- [x] I wrote at least basic e2e tests for the new functionalities or new paths
- [x] If this PR introduces UI changes, I manually tested this PR on iPhone 8 (iOS 10, 12, 14), iPhone 12, and iPhone 12 Max
- [x] If this PR modifies the behavior of an API, I did my best to make my new API backward compatible
- [x] If this PR modifies the API, I attempted to implement the feature with an example

_Documentation_

- [x] If this PR modifies the API, I updated the API Reference.
- [x] If the PR modifies the API, I wrote some documentation to deprecate it

# After merging


Debug App/Sources/View Controllers/New UI/MerchantHeadlessCheckoutKlarnaViewController.swift
```diff
@@ -0,0 +1,605 @@
...
+            
+            switch step {
+            case .paymentSessionAuthorized(let authToken):
+                showAlert(title: "Success", message: "Payment session completed with token: \(authToken)") { [unowned self] in
+                    navigationController?.popToRootViewController(animated: true)
+                }
+                
+            case .paymentSessionAuthorizationFailed:
+                showAlert(title: "Authorization", message: "Payment authorization failed") { [unowned self] in
+                    klarnaSessionAuthorizationComponent.reauthorizeSession()
```
NQuinn27, line 573: I dont think we can call reauthorize here, According to the klarna docs, we should use reauthorize when changing an already authorized session, it doesnt work to create a new one.

NQuinn27:
Some observations:

It looks like tokenization is missing from this PR, see `KlarnaTokenizationViewModel` and how it interacts with `PrimerKlarnaViewController` in Drop-in.

We are missing some control/flows to allow us to full test all of the possible flows in the Debug App:
- Manual finalization - we should add a toggle to test manual finalization, so we can see how that behaves.
- Reauthorization - klarna docs say this is for the case where the merchant wants to change something about the order and reauthorize. If we want to offer this functionality we need a way to test it somehow.

ikhrypunov:
Manual finalization step requires not just changes from iOS side, but also some additional payment session configurations. The session should have multi-step checkout flow, so that finalization should be required.

PR #730, opened 2023-11-08
chore: Add public settings for configuring supported card networks 

jnewc:
CHKT-1854

# What this PR does

Adds a new public setting for configuring which card networks are supported by a merchant.

* Apple Pay now also uses the configured networks too
  * Previously Apple Pay allowed _all_ networks regardless of whether a merchant supported them or not
  * The merchant must still configure `applePayOptions?.checkProvidedNetworks` to be true for networks to be checked
* Some renaming of models included to align with Android / improve naming
 
# Before merging

_QA_


Sources/PrimerSDK/Classes/PCI/Checkout Components/PrimerRawCardDataTokenizationBuilder.swift
```diff
@@ -60,6 +60,14 @@ class PrimerRawCardDataTokenizationBuilder: PrimerRawDataTokenizationBuilderProt
         }
     }
     
+    
+    /// Indicates whether or not one of the co-badged networks is present in the merchant config
+    /// If it is present, the remote card validation service will be called during card validation
+    var isCoBadgedCardsEnabled: Bool {
+        let supportedCardNetworks: [CardNetwork] = .supportedCardNetworks
+        return supportedCardNetworks.contains{ CardNetwork.coBadgedNetworks.contains($0) }
```
NQuinn27, line 68: TIL something. This syntax is a brainteaser... but nice usage 🙃

Not knowing this predicate-based contains, I probably would have ended up on something like `!Set(supportedCardNetworks).intersection(Set(CardNetwork.coBadgedNetworks)).isEmpty` but isn't a whole lot easier to grok in the end...

Not suggesting a change btw

jnewc: This is similar to what Andre suggested. Yeah, sets aren't that widely used ... but doesn't mean we can't though. In fact we could make the settings field type `Set<CardNetwork>` too to avoid duplication and make this a cleaner comparison.

I'll try a few things and see how it looks.

> This syntax is a brainteaser...

This is good enough reason to change it for me, I already don't like this syntax 😅 I've used an extension on Arrays to add a `contains(oneOf: ...)` method previously, but it seems overkill for this situation.

Sources/PrimerSDK/Classes/Core/BIN Data/CardValidationService.swift
```diff
@@ -71,7 +74,7 @@ class DefaultCardValidationService: CardValidationService, LogReporter {
                                                   forCardState: cardState)
         }.catch { error in
             // JN TODO: CHKT-1772 - send event
-            self.logger.error(message: "Error occurred while remotely validation '\(cardState.cardNumber)': \(error.localizedDescription)")
+            self.logger.error(message: "Error occurred while remote card validation: \(error.localizedDescription)")
```
semirp, line 77: Should this be a warning?

semirp (approved):

Sources/PrimerSDK/Classes/PCI/Checkout Components/PrimerRawCardDataTokenizationBuilder.swift
jnewc: I changed the config to expect a `Set` and it works great with array literals 🎉

Sources/PrimerSDK/Classes/Core/BIN Data/CardValidationService.swift
jnewc: Good question. I think it's an error because we can't display the network and, potentially (TBC), it means that a payment can't be completed.

What are your thoughts?

(Also, the grammar is terrible here - will update)

borisprimer (approved):

Sources/PrimerSDK/Classes/Core/BIN Data/CardValidationService.swift
semirp: Yeah if we won't fallback to local validation then this should be an error. If we are having a fallback it should be a warning as we are not breaking the flow.

jnewc: I'm 50/50 because this specific thing doesn't break the flow, but you won't be able to continue if we can't get the only supported network (e.g. CB is configured, but the API fails, so we can't complete payment)

NQuinn27 (approved):

PR #736, opened 2023-11-13
fix: Fix state machine to allow going back in steps

borisprimer:
This PR fixes a bug with Nol state machine. If a user wants to change the phone number after sending it for validation an error is returned. `submit()` was handling wrong step.

[JIRA](https://primerapi.atlassian.net/browse/CHKT-1928)
[SLACK](https://primerapi.slack.com/archives/C05DJ5LV309/p1699871603945539)


jnewc (approved):

NQuinn27:
Do we have tests for this state machine?

borisprimer:
> Do we have unit tests for this state machine?

We do now :)

NQuinn27 (approved):

PR #739, opened 2023-11-14
chore: Constrain remote validation logic to eight-digit BIN only

jnewc:
CHKT-1885

# What this PR does

Previously we triggered debounced validation on 6-8 characters, now we will trigger it:
1. When less than eight digits are entered, on eighth digit being entered 
2. After eight digits have been entered, only if the first eight digits change.

* The first request is no longer debounced to ensure that we make an early request for the most common use case (i.e. straight entry without user error)

# Before merging

_QA_


borisprimer (approved):
LGTM!

PR #744, opened 2023-11-15
chore: Deprecate CardComponentsManager

jnewc:
DEX-

# What this PR does

Deprecates CardComponentsManager and removes it from the debug app

* We no longer support this and will be removing all references in the documentations shortly.
* There is an `InternalCardComponentsManager` that is similar to `CardComponentsManager` but is out of scope for these changes (and is not publicly exposed)

# Before merging

_QA_

_Documentation_


jnewc:
@NQuinn27 @BorisNikolic should we use chore or fix for this?

NQuinn27:
Looks good! Two questions:
1 - Should we maintain the Debug App implementation as long as we provide the functionality, to verify we dont break it.
2 - Related, do we test CardComponents currently? (Thinking will any UI tests break, worth running Appium locally if you have not yet)

jnewc:
@NQuinn27 
1 - We can, although I think once it's marked as deprecated it should be considered "frozen" too, i.e. let's not make changes hereon because any hypothetical bug report should be met with a request to migrate to headless.
2 - I did look at the specs and didn't see any, nor any references to it. Happy to run do a locally run to verify 👍

NQuinn27 (approved):

PR #745, opened 2023-11-15
chore: update ApplePay presentation

NQuinn27:
This PR migrates from `PKPaymentAuthorizationViewController` -> `PKPaymentAuthorizationController`. The key change here is that the `.present()` function on `PKPaymentAuthorizationController` can fail, so we can inform the developer that there is an issue.

- Added a new error to be displayed here:
`PrimerError.unableToPresentApplePay`
`Recovery Suggestion`: "PassKit was unable to present the Apple Pay UI. Check merchantIdentifier and other parameters are set correctly for the current environment."

- Logs the error to the console
Screenshot 2023-11-15 at 14 19 10

**To Test**
To test this failure logic:
- Feed in an incorrect `merchantIdentifier` (Modify `ApplePayTokenizationViewModel.swift:l209`)
- Run the Debug app on a physical device (Simulator always succeeds 🤷) 
- Set currency to `GBP` and country to `GB`
- Trigger ApplePay via Drop-in UI


jnewc (approved):

borisprimer (approved):
I doublechecked behaviour of canMakePayment and canMakePayment(usingNetworks, it is the same for PKPaymentAuthorizationViewController and PKPaymentAuthorizationController. All good!

NQuinn27:
@jnewc I was able to complete end-end tests on Staging yesterday, everything working as expected. 

As your approval is still valid after merging master in, Ill merge later today if there are no objections 🚀

PR #746, opened 2023-11-16
chore: Add SwiftLint as a Run Phase

borisprimer:
Make SwiftLint active as a RunPhase, disable rules, fix some warnings.


NQuinn27 (approved):

Sources/PrimerSDK/Classes/Core/Logging/PrimerLogger.swift
```diff
@@ -7,7 +7,7 @@ public struct PrimerLogMetadata {
     public let file: String
     public let line: Int
     public let function: String
-
+
```
jnewc, line 10: Where are these spaces coming from? Is this something we can catch with 🥁 SwiftLint? 🥳

jnewc (approved):

semirp (approved):

PR #788, opened 2024-01-26
Release 2.20.1

primer-security-integrations:
Bumped 2.20.0 -> 2.20.1

### DIFF
--- a/.github/workflows/xcode-versions-build.yml
+++ b/.github/workflows/xcode-versions-build.yml
@@ -16,7 +16,7 @@ jobs:
       max-parallel: 3
       matrix:
         variant: [cocoapods, spm]
-        xcode: [14.3.1, 14.2, 14.1]
+        xcode: [14.3.1]
     steps:
       - name: Git - Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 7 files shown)


.github/workflows/xcode-versions-build.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       max-parallel: 3
       matrix:
         variant: [cocoapods, spm]
-        xcode: [14.3.1, 14.2, 14.1]
```
semirp, line 19: [q]: does this mean merchants won't be able ti integrate using these old-ish Xcode versions?

NQuinn27: No, just that we are tesing only on 14.3.1, it should be enough to say it "works on Xcode 14" since these are all point releases of the same Xcode version.
Thoughts @jnewc @BorisNikolic (this was an old suggestion about wasted build times)

borisprimer: I don't see a reason not to do it. Xcode 15 was released on September 18th. 4 months ago... And the diff between 14.3.1 and 14.2 and 14.1 is very minor.

borisprimer (approved):

.github/workflows/xcode-versions-build.yml
jnewc: Agree that 14.3.1 is enough 👍

jnewc (approved):

PR #810, opened 2024-02-27
refactor: NativeUIManager Composable Architecture

NQuinn27:
# Description

As discussed last week, this is a light refactor to bring NativeUIManager in line with the Composable Architecture.

Whats done:
- Break NativeUIManager into components to allow future APMs to be added
- Introduced NativeUIValidateable to handle common validation code + allow for individual custom validation

What is not done:
- Presentation of ApplePay and PayPal is not decoupled from PrimerInternal at this time


Sources/PrimerSDK/Classes/Core/PrimerHeadlessUniversalCheckout/Composable/NativeUI/PayPal/PayPalValidationComponent.swift
```diff
@@ -0,0 +1,25 @@
...
+//
+
+import Foundation
+
+struct PayPalValidationComponent: NativeUIValidateable {
+    let paymentMethodType = PrimerPaymentMethodType.payPal.rawValue
+
+    func validatePaymentMethod() throws {
+        if PrimerSettings.current.paymentMethodOptions.urlScheme == nil {
+            let err = PrimerError.invalidUrlScheme(urlScheme: nil,
```
borisprimer, line 15: I think we should decide whether we store errors as `err` or `error`, I vote for `error`. WDYT?

borisprimer (approved):
LGTM! One small nitpick comment regarding naming...

Also, I am not a big fan of git history pollution with commits without additional context, like for example `interim`. I am doing amend for all the `temporary` commits, I advice you do the same. Even squashing should be fine as long as we hove meaningful git commit messages ❤️

NQuinn27:
> LGTM! One small nitpick comment regarding naming...
> 
> Also, I am not a big fan of git history pollution with commits without additional context, like for example `interim`. I am doing amend for all the `temporary` commits, I advice you do the same. Even squashing should be fine as long as we hove meaningful git commit messages ❤️

Yep I agree, will keep it in mind - in this case I needed to quickly switch context and reached for a commit instead of stash 🙃

PR #837, opened 2024-04-11
Fix Nol pay Xcode errors

borisprimer:
[Lime was complaining](https://primerapi.slack.com/archives/C05DJ5LV309/p1712683167581349) about some funky errors in their CI/CD pipeline, this should solve it.


Sources/PrimerSDK/Classes/Core/PrimerHeadlessUniversalCheckout/Composable/NolPay/NolPayLinkedCardsComponent.swift
```diff
@@ -85,6 +87,8 @@ public class NolPayLinkedCardsComponent {
         isDebug =  PrimerLogging.shared.logger.logLevel == .debug
         #endif
 
+        #if canImport(PrimerNolPaySDK)
+
         guard nolPay == nil
         else {
             completion(.success(()))
```
NQuinn27, line 94: I know this didnt change this time around, but this is confusing to me.

In what case do we derive success from nolPay being non-nil?

borisprimer: If the start() function was already called before and you are calling it again, meaning you made a developer's mistake.

borisprimer: If the start() function was already called before and you are calling it again, meaning you made a developer's mistake.

borisprimer: If the start() function was already called before and you are calling it again. In this case, we have only one public getLinkedCardsFor that can be called multiple times. Only the first time you call the start() function you will init the nol SDK, every next time you will get success in the callback and you can continue with fetching the cards. Makes sense?

NQuinn27: Ok, makes sense now. 

Not for now, but if it is a developer mistake we should probably emit a debug message here to inform the developer that it was already initialised. Also serves to inform them that calling it again wont re-initialise.

NQuinn27 (approved):

PR #873, opened 2024-05-23
test: Tokenization view model unit tests (pt. 1)

jnewc:
# Description

Tests for:
* WebRedirectPaymentMethodTokenizationViewModel
* ApplePayTokenizationViewModel

# Other Notes

- Other changes that are not specifically related to the intent of the PR

# Manual Testing

_Add manual testing notes here if applicable, otherwise remove this section_

# Screenshots

_If applicable, otherwise remove this section_

# Contributor Checklist

# Reviewer Checklist

# Before Merging

# Other Stuff

### DIFF
--- a/Sources/PrimerSDK/Classes/PCI/Tokenization View Models/FormsTokenizationViewModel/CardFormPaymentMethodTokenizationViewModel.swift
+++ b/Sources/PrimerSDK/Classes/PCI/Tokenization View Models/FormsTokenizationViewModel/CardFormPaymentMethodTokenizationViewModel.swift
@@ -29,7 +29,7 @@ class CardFormPaymentMethodTokenizationViewModel: PaymentMethodTokenizationViewM
             paymentMethodType: self.config.type,
             isRequiringCVVInput: isRequiringCVVInput
         )
-        cardComponentsManager.delegate = self
+        manager.delegate = self
         return manager
     }()
 


Sources/PrimerSDK/Classes/User Interface/ApplePayPresentationManager.swift
```diff
@@ -0,0 +1,100 @@
...
+//  ApplePayPresentationManager.swift
+//  PrimerSDK
+//
+//  Created by Jack Newcombe on 23/05/2024.
+//
+
+import Foundation
+import PassKit
+
+protocol ApplePayPresenting {
```
borisprimer, line 11: Do we have a strict naming convention that we follow? I would rather write ApplePayPresentationProtocol than use the present continuous tense in names. I would always add a Protocol suffix because I could distinguish between classes and interfaces when I read code. Just a thought... We should have our public API assessed because different naming conventions are used, for example, `Primer3DSSDKProviderProtocol` and `protocol BanksComponentErrorable`, which is also a delegate btw, so why not `BanksComponentErrorDelegate`... Now I am not strictly advocating any of the ideas, I would just rather make one to rule them all. WDYT? @NQuinn27 @jnewc 
We are going to release a new version 3.0 at one point think it is ok to have a breaking change that is just a different API. BTW some of the protocol names are the same as on Android but also very different in some cases.

Screenshot 2024-05-23 at 23 12 17

borisprimer (approved):
One comment just to spark a conversation about naming convention subject.

PS: The work similar to what you did here is waiting on us in RN codebase. I am now picking up the low hanging fruits to try to pump the numbers, but some of the stuff is currently untestable without a refactor.

Sources/PrimerSDK/Classes/User Interface/ApplePayPresentationManager.swift
NQuinn27: Agree, we are not consistent here. 
My preference is to not have `Protocol` in the protocol name, but rather be descriptive on what the protocol does. As @jnewc mentioned elsewhere, this is worth a decision and a standard put in place.

NQuinn27 (approved):
LGTM - Might queue up some manual regression tests before we release to be doubly safe, wdyt?

PR #893, opened 2024-06-06
Quick fix to hide NOL pay from Drop-In

borisprimer:
I raised this PR just to pinpoint the problem and to fix it quickly, but there should be a better way to do it. 

I also created [this](https://primerapi.atlassian.net/browse/ACC-3382) JIRA ticket with findings.


NQuinn27:
Nice quick find.

On `PrimerUniversalHeadlessCheckout` we have `unsupportedPaymentMethodTypes`.
I would advocate for moving this and the new logic in Drop-in to `PrimerPaymentMethodType`

```
internal enum PrimerPaymentMethodType {
    ...
    
    var supportsHeadless: Bool {
        switch self {
            ...
        }
    }
    
    var supportsDropIn: Bool {
        switch self {
            ...
        }
    }
```

WDYT. Its not much more work and I think it could be the end state here. Simple and consolidated to one place.

PR #918, opened 2024-06-27
chore: Implement new workflow for updating of the translations from Phrase.com

borisprimer:


.github/workflows/manual-translation-update.yml
```diff
@@ -12,33 +11,40 @@ on:
...
+          PHRASE_ACCESS_KEY: ${{ secrets.PHRASE_ACCESS_KEY }}
+          PHRASE_PROJECT_ID: ${{ secrets.PHRASE_PROJECT_ID }}
 
       - name: Download Translations from Phrase
-        run: phrase pull --project-id ${{ secrets.PHRASE_PROJECT_ID }} --branch ${{ github.event.inputs.branch }}
+        run: phrase pull
 
       - name: Commit and Push Changes
         run: |
           git config --global user.email "action@github.com"
```
NQuinn27, line 42: We dont need the --global here or with user.name `git config user.email action@github.com` will do - also no need for the quote marks (although is probably fine with them).

.gitattributes
```diff
@@ -1 +1,2 @@
-*.strings diff
+*.strings text
+*.strings diff
```
borisprimer, line 2: @NQuinn27 @jnewc @semirp what is this red circle with a dash?

NQuinn27: It's applying the effect to all files of type string - `*.string`

https://git-scm.com/docs/gitattributes

NQuinn27 (approved):
Looks good

.gitattributes
borisprimer: I was asking about the UI here in GitHub, if you open https://github.com/primer-io/primer-sdk-ios/pull/918/files you will see it.

borisprimer: Screenshot 2024-06-28 at 14 52 50

PR #950, opened 2024-07-23
fix: Concurrent dependency access

NQuinn27:
# Description

Synchronizes dependency access to a queue. I was able to recreate Dependency access crashes with multiple concurrent accesses via Unit Tests. These changes aim to make it safely possible to access the Dependencies from multiple threads.


borisprimer (approved):
LGTM

Package.swift
```diff
@@ -14,9 +14,15 @@ let package = Package(
             targets: ["PrimerSDK"]
         )
     ],
+    dependencies: [
```
NQuinn27, line 17: These changes are the previously agreed changes to unblock their build. Not directly related

Sources/PrimerSDK/Classes/Core/Primer/DependencyInjection.swift
```diff
@@ -36,36 +40,47 @@ final internal class DependencyContainer {
...
+
+        if let dependency = Self.queue.sync(execute: { dependencies[key] as? T }) {
+            return dependency
         }
 
-        precondition(
-            dependency != nil,
-            "No dependency found for \(key)! must register a dependency before resolve."
-        )
+        return Self.queue.sync {
```
borisprimer, line 55: We could use only one queue.sync here.

borisprimer:
```
return Self.queue.sync {
            if let dependency = self.dependencies[key] as? T {
                return dependency
            }

            let dependency: T? = self.createDependency(for: key)
            if let dependency = dependency {
                self.dependencies[key] = dependency as AnyObject
                return dependency
            }

            preconditionFailure("No dependency found for \(key)! Must register a dependency before resolve.")
        }
```
something like this?

NQuinn27: Did you try to run Unit Tests with this change? It was crashing for me (deadlock, I think) with one queue.
Trying once more to make sure it wasnt a red herring

Sources/PrimerSDK/Classes/Core/Primer/DependencyInjection.swift
```diff
@@ -36,36 +40,47 @@ final internal class DependencyContainer {
...
+            if let dependency = dependency {
+                self.dependencies[key] = dependency as AnyObject
+            }
+
+            precondition(
+                dependency != nil,
+                "No dependency found for \(key)! must register a dependency before resolve."
+            )
+
+            return dependency!
```
borisprimer, line 70: force unwrap alert

NQuinn27: Yep, I dont like it either. But it existed before (red line 69), and the precondition should protect us.
To make this return a `T?` would be quite ugly at the callsites I suppose 😔

borisprimer: I see it is safe because of the precondition and it fails anyway, and we want it to fail here, I just don't like force unwrapping. `preconditionFailure` looks like a better solution. WDYT?

Sources/PrimerSDK/Classes/Core/Primer/DependencyInjection.swift
borisprimer: I did not try tests... leave it as it is, it is a nitpick anyway.

PR #976, opened 2024-08-19
chore: Add new header field to every API call

borisprimer:
[JIRA](https://primerapi.atlassian.net/browse/ACC-3929)

Is this the right place for adding this header?

### DIFF
--- a/Sources/PrimerSDK/Classes/PCI/Services/DefaultNetworkService.swift
+++ b/Sources/PrimerSDK/Classes/PCI/Services/DefaultNetworkService.swift
@@ -59,11 +59,12 @@ class DefaultNetworkService: NetworkService, LogReporter {
     func request<T: Decodable>(_ endpoint: Endpoint,
                                completion: @escaping ResponseCompletion<T>) -> PrimerCancellable? {
         do {
-            let request = try requestFactory.request(for: endpoint)
+            let identifier = String.randomString(length: 32)
+
+            let request = try requestFactory.request(for: endpoint, identifier: identifier)
 
             reportingService.report(eventType: .networkConnectivity(endpoint: endpoint))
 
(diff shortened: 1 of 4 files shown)


Sources/PrimerSDK/Classes/PCI/Services/DefaultNetworkService.swift
```diff
@@ -59,11 +59,14 @@ class DefaultNetworkService: NetworkService, LogReporter {
...
                                completion: @escaping ResponseCompletion<T>) -> PrimerCancellable? {
         do {
-            let request = try requestFactory.request(for: endpoint)
+            var request = try requestFactory.request(for: endpoint)
 
             reportingService.report(eventType: .networkConnectivity(endpoint: endpoint))
 
             let identifier = String.randomString(length: 32)
+            
+            request.addValue(identifier, forHTTPHeaderField: "X-Request-ID")
```
NQuinn27, line 68: This makes sense. My only suggestion would be to encapsulate this closer to the Request.

The options I see:
- Inject the `identifier` into `request(for:, identifier:)` and set the header in there.
- something like `func setIdentifier(for request: URLRequest)` in `NetworkRequestFactory.swift`
- something in an `extension URLRequest` in the same file.

First is probably the cleanest, just requires generating the random ID earlier. WDYT?

borisprimer: On it, good idea!

NQuinn27: These changes look good, so resolving this. Seems to be other issues with Unit Tests to resolve - other than those I am good to approve

NQuinn27 (approved):

PR #997, opened 2024-09-13
fix: Fix error reporting for apple pay display failure

jnewc:
# Description

There are currently two checks for whether Apple Pay can be presented, one in inside `ApplePayPresentationManager` and the other outside wrapping the call to `present` on the manager.

Only the former reports a presentation error on failure to show the Apple Pay dialog. This is never reached due to the outer wrap.

This PR removes the inner check and reports the error if the remaining check fails.

# Contributor Checklist

# Reviewer Checklist

# Before Merging

# Other Stuff

### DIFF
--- a/Sources/PrimerSDK/Classes/User Interface/ApplePayPresentationManager.swift
+++ b/Sources/PrimerSDK/Classes/User Interface/ApplePayPresentationManager.swift
@@ -10,6 +10,7 @@ import PassKit
 
 protocol ApplePayPresenting {
     var isPresentable: Bool { get }
+    var errorForDisplay: Error { get }
     func present(withRequest applePayRequest: ApplePayRequest,
                  delegate: PKPaymentAuthorizationControllerDelegate) -> Promise<Void>
 }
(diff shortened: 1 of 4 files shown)


Sources/PrimerSDK/Classes/User Interface/ApplePayPresentationManager.swift
```diff
@@ -33,13 +34,6 @@ class ApplePayPresentationManager: ApplePayPresenting, LogReporter {
...
-            guard isPresentable else {
-                let error = errorForDisplay
-                ErrorHandler.handle(error: error)
-                seal.reject(error)
-                return
-            }
-
             let request = createRequest(for: applePayRequest)
 
             let paymentController = PKPaymentAuthorizationController(paymentRequest: request)
```
NQuinn27, line 39: Should we this opportunity to return a different error below when `success == false`
This way we could more easily see if we are failing in the isPresentable check, or when we actually try to present.

NQuinn27: We do add extra info in the `errorForDisplay` - so we should be able to tell the difference. But afaict that (along with the line number) is not making it to Elastic 🤔

jnewc: The error that triggers on `success == false` is a slightly different error: `unableToPresentApplePay` rather than `unableToPresentPaymentMethod`.

NQuinn27 (approved):

PR #1024, opened 2024-10-14
fix: Move the cursor to the end of the text after pasting the card number

borisprimer:
# Description

[CHKT-4055
](https://primerapi.atlassian.net/browse/ACC-4055)

### DIFF
--- a/Sources/PrimerSDK/Classes/PCI/User Interface/Text Fields/PrimerCardNumberFieldView.swift
+++ b/Sources/PrimerSDK/Classes/PCI/User Interface/Text Fields/PrimerCardNumberFieldView.swift
@@ -87,6 +87,21 @@ public final class PrimerCardNumberFieldView: PrimerTextFieldView {
         }
 
         primerTextField.text = newText.withoutWhiteSpace.separate(every: 4, with: " ")
+
+        // Detect pasting of the card number
+        if string.count > 1 && (primerTextField.internalText?.count ?? 0) > 1 {
+            if let text = primerTextField.internalText {
+
+                // Get the position of the last character in the string
+                if let endPosition = primerTextField.position(from: primerTextField.beginningOfDocument, offset: text.count) {
+                    DispatchQueue.main.async {
+                        // Create a UITextRange from the endPosition to endPosition (for placing the cursor at the end)
+                        primerTextField.selectedTextRange = primerTextField.textRange(from: endPosition, to: endPosition)
+                    }
+                }
+            }
+        }
+
         return false
     }
 }
--- a/Tests/Primer/UI/PrimerCardNumberFieldViewTests.swift
+++ b/Tests/Primer/UI/PrimerCardNumberFieldViewTests.swift
@@ -116,4 +116,26 @@ final class PrimerCardNumberFieldViewTests: XCTestCase {
 
         waitForExpectations(timeout: 2.0)
     }
+
+    func testCursorMovesToEndAfterPasting() throws {
+        view.text = ""
+
+        // Simulate pasting a card number
+        _ = view.textField(view.textField,
+                           shouldChangeCharactersIn: NSRange(location: 0, length: 0),
+                           replacementString: "4242424242424242")
+
+        // Add a small delay to ensure the cursor movement code has executed
+        let expectation = XCTestExpectation(description: "Wait for cursor to move")
+
+        DispatchQueue.main.async {
+            // Check if the cursor is at the end of the text
+            let expectedPosition = self.view.textField.position(from: self.view.textField.endOfDocument, offset: 0)
+            XCTAssertEqual(self.view.textField.selectedTextRange?.start, expectedPosition)
+
+            expectation.fulfill()
+        }
+
+        wait(for: [expectation], timeout: 0.3)
+    }
 }


Sources/PrimerSDK/Classes/PCI/User Interface/Text Fields/PrimerCardNumberFieldView.swift
```diff
@@ -33,11 +33,32 @@ public final class PrimerCardNumberFieldView: PrimerTextFieldView {
...
+        // Restrict to 19 digits without spaces
         if string != "" && currentText.withoutWhiteSpace.count == 19 { return false }
+
+        // Get the new text after the replacement
         let newText = (currentText as NSString).replacingCharacters(in: range, with: string) as String
+
+        // Ensure the new text is numeric (except for spaces)
         if !newText.withoutWhiteSpace.isNumeric && !string.isEmpty { return false }
+
         primerTextField.internalText = newText
```
jnewc, line 46: Setting `text` below will also overwrite `internalText` so this isn't necessary

Sources/PrimerSDK/Classes/PCI/User Interface/Text Fields/PrimerCardNumberFieldView.swift
```diff
@@ -33,11 +33,32 @@ public final class PrimerCardNumberFieldView: PrimerTextFieldView {
...
+        // Format the text by adding spaces every 4 digits
+        let formattedText = newText.withoutWhiteSpace.separate(every: 4, with: " ")
+
+        // Set the formatted text back to the text field
+        primerTextField.text = formattedText
+
+        // Move the cursor to the last character of the text after pasting using Dispatch
+        DispatchQueue.main.async {
+            if let endPosition = primerTextField.position(from: primerTextField.endOfDocument, offset: 0) {
+                primerTextField.selectedTextRange = primerTextField.textRange(from: endPosition, to: endPosition)
```
jnewc, line 57: Won't this also move the cursor to the end if the user edits the middle of the card string? e.g. if they missed a few numbers

borisprimer: true. hmmm...

borisprimer:
Please review again! @jnewc @NQuinn27 I ended up with a hacky solution but it works!

Sources/PrimerSDK/Classes/PCI/User Interface/Text Fields/PrimerCardNumberFieldView.swift
```diff
@@ -87,6 +87,23 @@ public final class PrimerCardNumberFieldView: PrimerTextFieldView {
...
+
+        // Detect pasting of the card number
+        if string.count > 1 && (primerTextField.internalText?.count ?? 0) > 1 {
+            if let text = primerTextField.internalText {
+
+                // Get the position of the last character in the string
+                if let endPosition = primerTextField.position(from: primerTextField.beginningOfDocument, offset: text.count) {
+                    // Delaying it for 0.01 seconds because setting the selectedTextRange
+                    // before `shouldChangeCharactersIn:range` returns does not apply the value.
+                    DispatchQueue.main.asyncAfter(deadline: .now() + 0.01, execute: {
```
jnewc, line 99: If this is already on main queue (which it should be) can you just dispatch to main without the delay? That should make it happen as soon as this function exits

Sources/PrimerSDK/Classes/PCI/User Interface/Text Fields/PrimerCardNumberFieldView.swift
```diff
@@ -87,6 +87,23 @@ public final class PrimerCardNumberFieldView: PrimerTextFieldView {
...
+                        // Create a UITextRange from the endPosition to endPosition (for placing the cursor at the end)
+                        primerTextField.selectedTextRange = primerTextField.textRange(from: endPosition, to: endPosition)
```
jnewc, line 101: If the user pastes into the middle of the string what happens?

It's an edge case but I think having it move the cursor to the end of the text doesn't feel right. Could we move the cursor to the end of the inserted range instead?

Tests/Primer/UI/PrimerCardNumberFieldViewTests.swift
```diff
@@ -116,4 +116,27 @@ final class PrimerCardNumberFieldViewTests: XCTestCase {
...
+
+        // Simulate pasting a card number
+        _ = view.textField(view.textField,
+                           shouldChangeCharactersIn: NSRange(location: 0, length: 0),
+                           replacementString: "4242424242424242")
+
+        // Add a small delay to ensure the cursor movement code has executed
+        let expectation = XCTestExpectation(description: "Wait for cursor to move")
+
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.2) {
```
jnewc, line 131: If you switch to just a straight main dispatch above, you can also do that here

Sources/PrimerSDK/Classes/PCI/User Interface/Text Fields/PrimerCardNumberFieldView.swift
borisprimer: true, thanks!

PR #1025, opened 2024-10-14
fix: Fix Discover card network image not showing

borisprimer:
# Description

[CHKT-4054
](https://primerapi.atlassian.net/browse/ACC-4054)

Name of the icon in the code was wrong so UIImage was NIL


NQuinn27:
@BorisNikolic  I wonder if we can add some tests (Unit?) which ensure that these enums produce a valid image? Fine to ticket for later - but would be good to try ensure this cant happen again

PR #1031, opened 2024-10-21
fix: Prevent dismissal of Drop-in card form while a payment is active

borisprimer:
[CHKT-4038](https://primerapi.atlassian.net/browse/ACC-4038)

### DIFF
--- a/Sources/PrimerSDK/Classes/PCI/Tokenization View Models/FormsTokenizationViewModel/CardFormPaymentMethodTokenizationViewModel.swift
+++ b/Sources/PrimerSDK/Classes/PCI/Tokenization View Models/FormsTokenizationViewModel/CardFormPaymentMethodTokenizationViewModel.swift
@@ -328,7 +328,58 @@ class CardFormPaymentMethodTokenizationViewModel: PaymentMethodTokenizationViewM
             })
         }
 
-        super.start()
+        firstly {
+            self.startTokenizationFlow()
+        }
+        .done { paymentMethodTokenData in
+            self.paymentMethodTokenData = paymentMethodTokenData
+            self.processPaymentMethodTokenData()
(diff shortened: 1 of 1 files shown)


Sources/PrimerSDK/Classes/PCI/Tokenization View Models/FormsTokenizationViewModel/CardFormPaymentMethodTokenizationViewModel.swift
```diff
@@ -328,7 +328,58 @@ class CardFormPaymentMethodTokenizationViewModel: PaymentMethodTokenizationViewM
             })
         }
 
-        super.start()
```
NQuinn27, line 331: On the face of it I dont like introducing this repetition for the sake of one line difference. Understand a more elegant solution might not be simple, but were there any other approaches considered?

Would there be any way we could take `self.uiManager.primerRootViewController?.enableUserInteraction(true)` out of the `super` implementation, and inject it/execute it from the contexts it is needed.

Also begs the question: is `self.uiManager.primerRootViewController?.enableUserInteraction(true)` actually required in the `super` method? Will other flows break without this? @BorisNikolic

NQuinn27 (approved):

Sources/PrimerSDK/Classes/PCI/Tokenization View Models/FormsTokenizationViewModel/CardFormPaymentMethodTokenizationViewModel.swift
NQuinn27: Ticketed [here](https://primerapi.atlassian.net/browse/ACC-4433)

PR #1041, opened 2024-11-12
fix: Range or index out of bounds crash

borisprimer:
[JIRA](https://primerapi.atlassian.net/browse/ACC-4514)

PS: Also updated Cocoapods


NQuinn27:
Nice work.
Question: Do you have an idea what leads to the range being out of bounds in the first place?

borisprimer:
> Nice work. Question: Do you have an idea what leads to the range being out of bounds in the first place?

Operations that apply transformations (e.g., credit card number formatting, phone number grouping) can interfere with shouldChangeCharactersIn. For instance, if the code formats or adds characters programmatically (such as spaces in credit card numbers), this can cause unexpected range values. TBH I did not manage to reproduce it...

NQuinn27 (approved):

PR #1110, opened 2025-02-07
fix: Card Form for RTL languages

borisprimer:
[JIRA](https://primerapi.atlassian.net/browse/PRMSUP-3173)

The removed piece of code NEVER WORKED and only created this bug, which has been here for around 3 years.


NQuinn27 (approved):
Great catch!

PR #1202, opened 2025-06-28
refactor: Migrate PollingModule

OnurVar:


Tests/Primer/Modules/PollingModuleTests.swift
```diff
@@ -97,4 +139,29 @@ class PollingModuleTests: XCTestCase {
...
+        PollingModule.apiClient = mockApiClient
+        AppState.current.clientToken = nil
+
+        let pollingModule = PollingModule(url: URL(string: "https://random.url")!)
+
+        do {
+            _ = try await pollingModule.start()
+            XCTAssert(false, "Polling succeeded, but it should fail with error .invalidClientToken")
+        } catch {
+            if let primerErr = error as? PrimerError, case .invalidClientToken = primerErr {
```
henry-cooper-primer, line 160: suggestion: this might be more appropriate as a guard:
```suggestion
             guard let primerErr = error as? PrimerError, case .invalidClientToken = primerErr else {
                 XCTAssert(false, "Polling failed with error \(error.localizedDescription), but it should fail with error .invalidClientToken")
```

henry-cooper-primer (dismissed):

henry-cooper-primer (approved):

PR #1217, opened 2025-07-02
refactor: Migrate PrimerAPIConfigurationModule, VaultService, PrimerDelegate, PrimerHeadlessUniversalCheckout and VaultManager

OnurVar:


Sources/PrimerSDK/Classes/Core/PrimerHeadlessUniversalCheckout/Managers/VaultManager.swift
```diff
@@ -746,17 +1010,82 @@ extension PrimerHeadlessUniversalCheckout {
...
+                        err = NSError.emptyDescriptionError
+                    }
+                    throw err
+
+                case .succeed, .continueWithNewClientToken:
+                    return nil
+                }
+            } else if resumeDecision.type is PrimerHeadlessUniversalCheckoutResumeDecision.DecisionType {
+                self.paymentCheckoutData = nil
+                // TODO: REVIEW_CHECK - What should we return here?
```
henry-cooper-primer, line 1042:
```swift
else if resumeDecision.type is PrimerHeadlessUniversalCheckoutResumeDecision.DecisionType {
    // No need to continue if manually handling resume
    self.paymentCheckoutData = nil
```

The original promise above was left unresolved so its state was undefined. But returning nil _will_ continue the flow so I'm not sure this is the correct answer either. Might be one for @NQuinn27

NQuinn27 (approved):

PR #1301, opened 2025-08-21
feat: Add appearance mode override to respect app-level theme settings

borisprimer:
This PR addresses ESC-591 where apps with custom theme management (e.g., "Always Light" mode) experience unreadable UI elements when the system is in dark mode.

**Problem:**
  The SDK was using UIScreen.main.traitCollection.userInterfaceStyle to determine dark/light mode, which only reflects the system setting. This caused issues for apps like Dabble that force light
  mode while the device is in dark mode, resulting in dark backgrounds with dark text.

**Solution:**
  - Added PrimerAppearanceMode enum with .system, .light, and .dark options
  - Added appearanceMode property to PrimerUIOptions (defaults to .system for backward compatibility)
  - Updated UIScreen.isDarkModeEnabled and PrimerColor.dynamic to respect the override
  - Apps can now control SDK appearance independently of system settings

**Usage:**
  let uiOptions = PrimerUIOptions(
      appearanceMode: .light  // Force light mode
  )

  Testing:
  - ✅ Build succeeds without errors
  - ✅ Backward compatible - existing integrations work without changes
  - ✅ Resolves the reported issue where payment buttons were unreadable

Fixes: ESC-591, ACC-5794
  
[DOCS PR](https://gitlab.com/primer-io/general/public-docs/-/merge_requests/57)


Tests/Primer/Extensions/PrimerColorTests.swift
```diff
@@ -52,8 +52,77 @@ final class PrimerColorTests: XCTestCase {
...
+    // MARK: - Dynamic Color Tests with Appearance Mode
+    
+    func testDynamicColorWithSystemAppearance() {
+        // Given: System appearance mode
+        let settings = PrimerSettings(
+            uiOptions: PrimerUIOptions(appearanceMode: .system)
+        )
+        Primer.shared.configure(settings: settings)
+        
+        let lightColor = PrimerColor(red: 255, green: 255, blue: 255) // White
```
henry-cooper-primer, line 65: I think we can remove the comments (below too)

Debug App/Podfile.lock
henry-cooper-primer, line 1: can we remove this if not needed or do a separate PR if it is?

henry-cooper-primer (approved):

PR #1355, opened 2025-09-19
chore: Add SwiftFormat rules & commit hook trigger

henry-cooper-primer:
This pr serves to automate the process for adding a pre-commit hook that will run SwiftFormat on files using the default rules as well as opt-in ones that have been agreed (in process) by the iOS team. 

Simply run `make hook` once and the files changed will be formatted according to the rules defined. 

The pre-commit hook uses the [git-format-staged](https://github.com/hallettj/git-format-staged) script, which is the recommended method


borisprimer (approved):
The main consideration is coordinating the initial codebase reformatting. This will create a large commit and potentially cause merge conflicts with active branches. I suggest planning a coordinated "format day" when minimal feature branches are in flight to minimize disruption.

Should we consider keeping SwiftLint in a follow-up PR to complement formatting with static analysis and style rule?

Great job! 💪

NQuinn27 (approved):

PR #1649, opened 2026-03-27
chore: Update call sites for State Processor

henry-cooper-primer:


OnurVar (approved):
@henry-cooper-primer Should the `URLSession` be stored as a property so the cache persists across fetch calls? Right now a new session and cache are created each time, so nothing is actually cached.

henry-cooper-primer:
> @henry-cooper-primer Should the `URLSession` be stored as a property so the cache persists across fetch calls? Right now a new session and cache are created each time, so nothing is actually cached.

oops great spot. Thanks

PR #1688, opened 2026-04-14
feat: Implement ADYEN_SWISH in CheckoutComponents

borisprimer:
## Summary
- Register `ADYEN_SWISH` (Swish) as a WebRedirect payment method
- Backend sends `WEB_REDIRECT` on mobile via `mobile-override-implementation-type`, so it auto-registers with the existing filter
- Uses `swish-logo-colored` icon asset (already in bundle)

**Jira:** [ACC-7019](https://primerapi.atlassian.net/browse/ACC-7019)

## Changes
| File | Change |
|------|--------|
| `PrimerPaymentMethodType.swift` | Add `adyenSwish` enum case + provider |
| `PrimerPaymentMethodType+ImageName.swift` | Add `icon` (swish-logo-colored) |
| `UserInterfaceModule.swift` | Add to exhaustive switch (return nil) |
| `SwishRegistrationTests.swift` | **New** — 11 registration tests |
| `PrimerPaymentMethodTypeImageNameTests.swift` | Add icon test |

## Test plan
- [x] Build succeeds
- [x] All 12 new tests pass (11 registration + 1 image)
- [x] All existing tests pass
- [x] SwiftFormat + SwiftLint clean

[ACC-7019]: https://primerapi.atlassian.net/browse/ACC-7019?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/PrimerSDK/Classes/Modules/UserInterfaceModule.swift
+++ b/Sources/PrimerSDK/Classes/Modules/UserInterfaceModule.swift
@@ -756,7 +756,8 @@ final class UserInterfaceModule: NSObject, UserInterfaceModuleProtocol {
             return nil
         case .fintechtureSmartTransfer, .fintechtureImmediateTransfer:
             return nil
-        case .mollieGiftcard:
+        case .adyenSwish,
+             .mollieGiftcard:
             return nil
         }
     }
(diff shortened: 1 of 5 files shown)


Sources/PrimerSDK/Classes/Modules/UserInterfaceModule.swift
```diff
@@ -756,7 +756,8 @@ final class UserInterfaceModule: NSObject, UserInterfaceModuleProtocol {
             return nil
         case .fintechtureSmartTransfer, .fintechtureImmediateTransfer:
             return nil
-        case .mollieGiftcard:
+        case .adyenSwish,
```
nquinn-primer, line 759: Two things here:

- Why group adyenSwish with mollieGiftcard? We have other adyen methods at the beginning.
- Is nil the best default we can do here. Does `DisplayMetadata` in the configuration response return anything today which we could use as a local fallback.

PR #1693, opened 2026-04-15
fix: Pass correct paymentMethodType to QRCode interactor

borisprimer:
## Summary

**Jira**: ACC-7131

- The QRCode interactor was registered in DI with `paymentMethodType: ""`, causing all QR code payments (PromptPay, Xfers PayNow) to send an empty string to the backend
- Introduces `QRCodePaymentInteractorFactory` using the existing `Factory<Product, Params>` protocol so the correct payment method type is passed at scope-creation time
- This is the first usage of the Factory pattern in CheckoutComponents, leveraging the existing DI framework infrastructure

## Test plan

- [x] All 22 QRCode-related tests pass (QRCodePaymentMethodTests, ProcessQRCodePaymentInteractorTests, DefaultQRCodeScopeTests)
- [x] Build succeeds
- [ ] Manual: test with a real QR code payment method (e.g., PromptPay) to verify correct type is sent

### DIFF
--- a/Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Domain/Interactors/QRCodePaymentInteractorFactory.swift
+++ b/Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Domain/Interactors/QRCodePaymentInteractorFactory.swift
@@ -1,0 +1,21 @@
+//
+//  QRCodePaymentInteractorFactory.swift
+//
+//  Copyright © 2026 Primer API Ltd. All rights reserved. 
+//  Licensed under the MIT License. See LICENSE file in the project root for full license information.
+
+@available(iOS 15.0, *)
+struct QRCodePaymentInteractorFactory: Factory, @unchecked Sendable {
+  typealias Product = ProcessQRCodePaymentInteractor
+  typealias Params = String
+
+  private let repository: QRCodeRepository
+
+  init(repository: QRCodeRepository) {
+    self.repository = repository
+  }
+
+  func create(with paymentMethodType: String) async throws -> ProcessQRCodePaymentInteractor {
+    ProcessQRCodePaymentInteractorImpl(repository: repository, paymentMethodType: paymentMethodType)
+  }
+}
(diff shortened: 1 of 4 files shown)


Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Domain/Interactors/QRCodePaymentInteractorFactory.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  QRCodePaymentInteractorFactory.swift
+//
+//  Copyright © 2026 Primer API Ltd. All rights reserved. 
+//  Licensed under the MIT License. See LICENSE file in the project root for full license information.
+
+@available(iOS 15.0, *)
+struct QRCodePaymentInteractorFactory: Factory, @unchecked Sendable {
+  typealias Product = ProcessQRCodePaymentInteractor
```
henry-cooper-primer: both aliases unused?

Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Domain/Interactors/QRCodePaymentInteractorFactory.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  QRCodePaymentInteractorFactory.swift
+//
+//  Copyright © 2026 Primer API Ltd. All rights reserved. 
+//  Licensed under the MIT License. See LICENSE file in the project root for full license information.
+
+@available(iOS 15.0, *)
```
henry-cooper-primer, line 7: does this need to be @unchecked? What's error we're getting?

Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Domain/Interactors/QRCodePaymentInteractorFactory.swift
bthebladeprimer: Yes — removed. Swift infers `Product` and `Params` from the `create(with:)` signature.

Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Domain/Interactors/QRCodePaymentInteractorFactory.swift
bthebladeprimer: Turns out it wasn't needed — the struct only stores a `let repository`, and the compiler is fine with plain `Factory` conformance without any Sendable annotation. Removed.

nquinn-primer (approved):

PR #1705, opened 2026-04-20
fix: Propagate critical DI registration failures in ComposableContainer

borisprimer:
# Description

ACC-7133

`ComposableContainer.configure()` previously wrapped every registration in a `guardedRegister` that swallowed and logged all errors, and the only validation (`performHealthCheck()`) was gated behind `#if DEBUG` with no throwing behavior. That meant a broken container could be published successfully in release builds, and downstream `try?` resolves would fail silently — surfacing as confusing "nothing happens" bugs at payment time instead of a clear error at init.

**What changed:**

- `configure()` is now `async throws` and its single production caller (`CheckoutSDKInitializer.initialize()`, already `throws`) propagates the error naturally.
- Registration helpers split by intent:
  - `criticalRegister` — logs and **rethrows**, used for `PrimerSettings`, analytics service + interactor, `ConfigurationService`, `ValidationService`, `HeadlessRepository`, `PaymentMethodMapper`.
  - `guardedRegister` — keeps the existing log-and-swallow behavior for optional per-payment-method wiring so a single APM failing doesn't block checkout.
- New `validateCriticalDependencies()` runs in **every** build config after registration and resolves each critical type. This triggers the factory closures (which otherwise only run at resolve time) so misconfiguration surfaces immediately.
- `DIContainer.setContainer(...)` now runs **after** validation, so a partially-configured container is never published globally.
- The DEBUG-only `performHealthCheck()` diagnostic logging is unchanged.

**Not a breaking change** — `configure()` is not part of the public merchant API; the only caller was already in a throwing context.

# Manual Testing

Covered by existing + new automated tests. The failure path delegates to `Container.resolve(...)`, which is already extensively covered by `ContainerErrorTests` and `DIContainerTests`.

# Contributor Checklist

- [x] I have manually tested newly added UX

### DIFF
--- a/Sources/PrimerSDK/Classes/CheckoutComponents/Internal/DI/ComposableContainer.swift
+++ b/Sources/PrimerSDK/Classes/CheckoutComponents/Internal/DI/ComposableContainer.swift
@@ -17,12 +17,14 @@ final class ComposableContainer: LogReporter {
     self.settings = settings
   }
 
-  func configure() async {
-    await registerInfrastructure()
-    await registerValidation()
+  func configure() async throws {
+    try await registerInfrastructure()
+    try await registerValidation()
     await registerInteractors()
     await registerPaymentInteractors()
-    await registerData()
+    try await registerData()
+
+    try await validateCriticalDependencies()
 
     await DIContainer.setContainer(container)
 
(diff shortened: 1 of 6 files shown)


Sources/PrimerSDK/Classes/CheckoutComponents/Internal/DI/ComposableContainer.swift
```diff
@@ -297,6 +313,19 @@ extension ComposableContainer {
     }
   }
 
+  /// Runs in every build config. Resolving exercises the factory closures that
+  /// are deferred until resolution time, catching misconfiguration at init
+  /// instead of surfacing as a silent failure during payment.
+  fileprivate func validateCriticalDependencies() async throws {
```
henry-cooper-primer: can it not be private?

henry-cooper-primer (approved):

Sources/PrimerSDK/Classes/CheckoutComponents/Internal/DI/ComposableContainer.swift
bthebladeprimer: it is private now

PR #1734, opened 2026-05-07
refactor: Tighten payment-method protocol surface [ACC-7173]

bthebladeprimer:
# Description

[ACC-7173](https://primerapi.atlassian.net/browse/ACC-7173)

Final follow-up to ACC-7135 / ACC-7172 / ACC-7171. Cleans up the internal `PaymentMethodProtocol` machinery and adds public-protocol metatype access on `PrimerCheckoutScope` so most call sites no longer need the concrete `Default*Scope` type to look a scope up.

## Scope A — `PaymentMethodProtocol` cleanup

- Drop the `associatedtype ScopeType: PrimerPaymentMethodScope` from `PaymentMethodProtocol` and have `createScope(...)` return `any PrimerPaymentMethodScope` directly.
- Delete dead code that depended on `ScopeType`: the `content<V>` and `defaultContent` requirements (every conformer was a `fatalError` stub, no callers in `Sources/`, `Tests/`, or `Debug App/`), the unused `register<T: PaymentMethodProtocol>` overload, the `typeToIdentifier` field, and the unused `createScope<T>(_ scopeType:)` overload.
- Each PaymentMethod's `register()` now calls the existing low-level `register(forKey:scopeCreator:viewCreator:)` directly (the same pattern `WebRedirectPaymentMethod` was already using). Drops 8× `typealias ScopeType = Default*Scope` and 16 stub bodies.
- Removes invariant `paymentMethodType: String { get }` from `PrimerAdyenKlarnaScope` (always `"ADYEN_KLARNA"` — pure boilerplate).

## Scope B Approach 1 — per-protocol existential overloads on `PrimerCheckoutScope`

- Add 10 new `getPaymentMethodScope(_:)` overloads that take `(any Primer<Method>Scope).Type` so call sites can pass the public-protocol metatype instead of the concrete `Default*Scope`. The protocol extension provides `nil`-returning defaults so test mocks and merchant-side conformers don't have to stub all 10. `DefaultCheckoutScope` overrides each with the real cache-backed lookup via a private generic helper.
- Public-API addition only — no break.
- Rebind 7 of 10 internal call sites to the public-protocol metatype (Klarna, AdyenKlarna, Ach, PayPal, WebRedirect, QRCode, BillingAddressRedirect). 3 sites stay on the concrete metatype with inline `audit §2a` notes — `CardFormScreen` takes an internal sub-protocol, and `ApplePayScreen` / `FormRedirectContainerView` use `@ObservedObject Default*Scope` which SwiftUI requires to be concrete.

## Bonus

- Fix a public-API leak in `CardFormProvider` (a public type) where `getPaymentMethodScope(DefaultCardFormScope.self)` was reaching into the internal concrete class. Now uses `PrimerCardFormScope.self`.
- 4 test-suite cleanups — convert `if let scope = … as? Default*Scope { XCTAssertEqual(...) }` to `let scope = try XCTUnwrap(... as? Default*Scope); XCTAssertEqual(...)` in `WebRedirectPaymentMethodTests` and `QRCodePaymentMethodTests`.

The SDK is in beta with no merchants on this surface yet, so the public-API additions are safe.

# Manual Testing

- [ ] Run a card payment end-to-end (covers Card path)
- [ ] Run a Klarna payment if backend supports it (covers metatype rebind)

# Contributor Checklist

- [x] I have open a documentation PR, if applicable (N/A — internal refactor)

[ACC-7173]: https://primerapi.atlassian.net/browse/ACC-7173

### DIFF
--- a/Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Presentation/Scope/DefaultCheckoutScope.swift
+++ b/Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Presentation/Scope/DefaultCheckoutScope.swift
@@ -374,6 +374,57 @@ final class DefaultCheckoutScope: PrimerCheckoutScope, ObservableObject, LogRepo
     getPaymentMethodScope(for: methodType.rawValue)
   }
 
+  // MARK: - Per-protocol scope access (existential metatypes)
+
+  func getPaymentMethodScope(_ scopeType: (any PrimerCardFormScope).Type) -> (any PrimerCardFormScope)? {
+    findScope()
+  }
+
+  func getPaymentMethodScope(_ scopeType: (any PrimerKlarnaScope).Type) -> (any PrimerKlarnaScope)? {
+    findScope()
+  }
+
+  func getPaymentMethodScope(_ scopeType: (any PrimerAdyenKlarnaScope).Type) -> (any PrimerAdyenKlarnaScope)? {
+    findScope()
+  }
+
+  func getPaymentMethodScope(_ scopeType: (any PrimerWebRedirectScope).Type) -> (any PrimerWebRedirectScope)? {
+    findScope()
+  }
+
+  func getPaymentMethodScope(_ scopeType: (any PrimerFormRedirectScope).Type) -> (any PrimerFormRedirectScope)? {
+    findScope()
+  }
+
+  func getPaymentMethodScope(_ scopeType: (any PrimerBillingAddressRedirectScope).Type) -> (any PrimerBillingAddressRedirectScope)? {
+    findScope()
+  }
+
+  func getPaymentMethodScope(_ scopeType: (any PrimerApplePayScope).Type) -> (any PrimerApplePayScope)? {
+    findScope()
+  }
+
+  func getPaymentMethodScope(_ scopeType: (any PrimerPayPalScope).Type) -> (any PrimerPayPalScope)? {
+    findScope()
+  }
+
+  func getPaymentMethodScope(_ scopeType: (any PrimerQRCodeScope).Type) -> (any PrimerQRCodeScope)? {
+    findScope()
+  }
+
+  func getPaymentMethodScope(_ scopeType: (any PrimerAchScope).Type) -> (any PrimerAchScope)? {
+    findScope()
+  }
+
+  /// Returns the first cached scope conforming to the requested protocol existential.
+  private func findScope<P>() -> P? {
+    guard let scope = paymentMethodScopeCache.values.first(where: { $0 is P }) as? P else { return nil }
+    if let paymentMethodScope = scope as? any PrimerPaymentMethodScope {
+      currentPaymentMethodScope = paymentMethodScope
+    }
+    return scope
+  }
+
   func onDismiss() {
     updateState(.dismissed)
     updateNavigationState(.dismissed)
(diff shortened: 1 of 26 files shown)


Sources/PrimerSDK/Classes/CheckoutComponents/PaymentMethods/Ach/AchPaymentMethod.swift
```diff
@@ -102,19 +92,13 @@ extension AchPaymentMethod {
...
-      fatalError("Default content method should be implemented by the CheckoutComponents framework")
-    }
-
     @MainActor
     static func register() {
-      PaymentMethodRegistry.shared.register(TestAchPaymentMethod.self)
+      PaymentMethodRegistry.shared.register(
+        forKey: paymentMethodType,
+        scopeCreator: { try await createScope(checkoutScope: $0, diContainer: $1) },
+        viewCreator: { createView(checkoutScope: $0) }
```
henry-cooper-primer: `viewCreator: createView`. In numerous other spots too

Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Presentation/Scope/DefaultCheckoutScope.swift
```diff
@@ -374,6 +374,57 @@ final class DefaultCheckoutScope: PrimerCheckoutScope, ObservableObject, LogRepo
     getPaymentMethodScope(for: methodType.rawValue)
   }
 
+  // MARK: - Per-protocol scope access (existential metatypes)
+
+  func getPaymentMethodScope(_ scopeType: (any PrimerCardFormScope).Type) -> (any PrimerCardFormScope)? {
```
henry-cooper-primer: why are we passing in `scopeType`? How is being used?

bthebladeprimer:
Bundled a tiny extra commit (`f148ac947`) to fix a timing flake that surfaced after the umbrella merge:

**`test_submit_pollingStarted_transitionsToAwaitingExternalCompletion`** in `DefaultFormRedirectScopeTests` was intermittently failing because the test waits 200ms after `submit()` and asserts the scope is in the intermediate `.awaitingExternalCompletion` state — but the mock's `executeDelay` was 300ms, only a 100ms margin. On a loaded CI runner the test's `Task.sleep(200ms)` can stretch past the mock's 300ms sleep, the mock returns, status flips to `.success`, assertion reads `.success` and fails.

Fix is one line: bump `executeDelay` from `0.3` → `0.6` to give a ≥400ms margin. The production state-machine flow is unchanged — `DefaultFormRedirectScope` and the test file haven't been touched by ACC-7173, ACC-7172, or ACC-7171, so the flake is pre-existing and unrelated to this refactor.

Folded into this PR rather than a separate `chore:` to avoid CI ping-ponging across PRs.

Sources/PrimerSDK/Classes/CheckoutComponents/PaymentMethods/Ach/AchPaymentMethod.swift
bthebladeprimer: Done in c93b973d8 — switched all 7 simple PaymentMethod registrations (Card, ApplePay, PayPal, Klarna+TestKlarna, AdyenKlarna, Ach+TestAch, FormRedirect's Blik+MBWay) to direct function references. WebRedirect and BillingAddressRedirect were already using them.

Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Presentation/Scope/DefaultCheckoutScope.swift
bthebladeprimer: Good catch — `scopeType` is unused at runtime. It exists only as a type discriminator for overload resolution at the call site (Swift can't pick the right overload by return type alone). `findScope<P>()` infers `P` from the return type via generic inference. Renamed the internal name from `scopeType` → `_` in dc930090f and added a short comment above the block so it's obvious the param isn't read.

henry-cooper-primer (approved):

PR #1735, opened 2026-05-12
chore: Add @_spi to PrimerFoundation internals

henry-cooper-primer:


semirp (approved):

Debug App/Podfile
```diff
@@ -10,6 +10,9 @@ target 'Debug App' do
   # Pods for Debug App
   pod 'PrimerSDK', :path => '../'
   pod 'PrimerFoundation', :path => '../'
+  pod 'PrimerStepResolver', :path => '../'
```
nquinn-primer, line 13: Will merchants need to add multiple new pods, or are we banking on Cocoapods not being a thing by the time most are onboarding?

nquinn-primer: Also, it feels slightly off that these are both added at the integration side, and `@_spi` - could they be subspecs of a public pod and be hidden from the merchant entirely? 🤔

nquinn-primer (approved):
Looks good, a couple of questions which arent blocking

Debug App/Podfile
henry-cooper-primer: Sorry yeah I should have commented. This is just so build with Appetize doesn't fail at this point. Since changes are made to the SDK, but the corresponding changes aren't reflected on the trunk, the build fails, so this is mitigated by passing the path in this way that uses the pods locally.

Once pushed to trunk, I'll create a PR to remove the changes from the Podfile, à la #1720 . Everything will be transparent to merchants, they simply bring in PrimerSDK

PR #1759, opened 2026-05-27
chore: Move PrimerSettings objects to PrimerCore

henry-cooper-primer:


Modules/PrimerCore/Sources/PrimerCore/Settings/PrimerPaymentMethodOptions.swift
```diff
@@ -4,25 +4,20 @@
...
 
-public final class PrimerPaymentMethodOptions: PrimerPaymentMethodOptionsProtocol, Codable {
-
-    private let urlScheme: String?
-    let applePayOptions: PrimerApplePayOptions?
-    var klarnaOptions: PrimerKlarnaOptions?
+public final class PrimerPaymentMethodOptions: Codable {
 
+    public let urlScheme: String?
+    public let applePayOptions: PrimerApplePayOptions?
```
marjaimate: is that OK to make these public? In the diff they were private to `PrimerPaymentMethodOptions`

Modules/PrimerCore/Sources/PrimerCore/Settings/PrimerStripeOptions.swift
```diff
@@ -11,8 +11,8 @@ public final class PrimerStripeOptions: Codable {
         case templateMandate(merchantName: String)
     }
 
-    var publishableKey: String
-    var mandateData: MandateData?
+    public var publishableKey: String
```
marjaimate: Same as the comment on `PrimerPaymentMethodOptions` about making these public.

marjaimate (dismissed):

Modules/PrimerCore/Sources/PrimerCore/Settings/PrimerPaymentMethodOptions.swift
henry-cooper-primer: good shout. They do need to be public since the sdk needs to access them now they are in a different target. however they can be SPI gated (done now) to prevent merchants touching them

marjaimate (dismissed):

marjaimate (approved):

PR #1816, opened 2026-06-24
fix: Convert Apple Pay amounts by currency decimalDigits

bthebladeprimer:
# Description

ORC-7507

Apple Pay amount conversion divided **every** non-zero-decimal currency by `100`, instead of `10^decimalDigits`. For **3-decimal currencies** (BHD, IQD, JOD, KWD, LYD, OMR, TND) this divided by 100 instead of 1000, so the Apple Pay sheet displayed an amount **10× larger** than the real charge (e.g. `8000` KWD minor units → `80.00` instead of `8.000`).

This was found while investigating ESC-982 (Albanian Lek on Google Pay). Note: **ALL is not affected on iOS** — it is correctly a 2-decimal currency, and iOS has no Google Pay. ESC-982 is a Web-only fix; this is a separate latent iOS bug. Android already handles it correctly (`amount / 10^fractionDigits`).

### What these changes do

- Add a single source of truth `Currency.minorUnitDivisor` (`10^decimalDigits`).
- Use it in place of the binary `isZeroDecimal` (÷1 or ÷100) check across:
  - `ApplePayOrderItem` — payment summary item total
  - `Int.formattedCurrencyAmount(currency:)` — recurring / deferred / automatic-reload Apple Pay requests
  - `Int.toCurrencyString(currency:locale:)` — divisor **and** min/max fraction digits now use `decimalDigits`
  - `ApplePayTokenizationViewModel` — shipping method amounts
- Add `KWD_CurrencyFormattingTests` (3-decimal) covering the conversion and 0/2-decimal regression.

No breaking changes. Amounts stored in the dashboard and sent to the PSP are unaffected — this is display/formatting only.

# Other Notes

- `Int.formattedCurrencyAmount` previously built a `NumberFormatter` it never used; removed.
- `IPay88TokenizationViewModel` has a separate hardcoded `/100`, but it is currently harmless (MYR is 2-decimal) and out of scope here.
- The `internal`→default, `&&`→`,` and copyright-year changes in `ApplePayOrderItem` are SwiftFormat/SwiftLint auto-fixes.

# Manual Testing

Covered by unit tests (`KWD_CurrencyFormattingTests` + existing per-currency suites). No UI change beyond the corrected amount string.

# Contributor Checklist

### DIFF
--- a/Modules/PrimerFoundation/Sources/PrimerFoundation/Model/Currency.swift
+++ b/Modules/PrimerFoundation/Sources/PrimerFoundation/Model/Currency.swift
@@ -10,17 +10,17 @@ import Foundation
 public struct Currency: Codable {
     public let code: String
     public let decimalDigits: Int
-    
+
     private enum CodingKeys: String, CodingKey {
         case code = "c"
         case decimalDigits = "m"
     }
-    
+
     public init(code: String, decimalDigits: Int) {
         self.code = code
         self.decimalDigits = decimalDigits
     }
-    
+
     public var symbol: String? {
         let localeIdentifier = Locale.identifier(fromComponents: [NSLocale.Key.currencyCode.rawValue: code])
         let locale = Locale(identifier: localeIdentifier)
@@ -29,5 +29,10 @@ public struct Currency: Codable {
 
     public var isZeroDecimal: Bool {
         decimalDigits == 0
+    }
+
+    /// Divisor that converts a minor-unit amount into its major-unit value (`10^decimalDigits`).
+    public var minorUnitDivisor: Decimal {
+        pow(10, decimalDigits)
     }
 }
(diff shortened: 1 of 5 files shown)


Modules/PrimerFoundation/Sources/PrimerFoundation/Model/Currency.swift
```diff
@@ -30,4 +30,9 @@ public struct Currency: Codable {
     public var isZeroDecimal: Bool {
```
nquinn-primer: If this is no longer used we could remove it?

nquinn-primer (dismissed):
Looks good, just one comment

Modules/PrimerFoundation/Sources/PrimerFoundation/Model/Currency.swift
bthebladeprimer: removed

nquinn-primer (approved):

PR #1838, opened 2026-07-23
feat: Introduce PrimerBDCUI non-ui Objects

henry-cooper-primer:


Modules/PrimerBDCUI/Sources/PrimerUI/Classes/Router.swift
```diff
@@ -0,0 +1,80 @@
...
+	
+    func pop(_ value: Int = 1) { if !path.isEmpty { path.removeLast(value) } }
+
+    func popToRoot() { pop(steps.count) }
+
+    private func publish() { objectWillChange.send() }
+}
+
+extension UINavigationController: @retroactive UIGestureRecognizerDelegate {
+    override open func viewDidLoad() {
```
OnurVar: this overrides viewDidLoad on UINavigationController itself, so it hits every nav controller in the merchant's app, not just ours. maybe we can put it on PrimerNavigationController?

Modules/PrimerBDCUI/Sources/PrimerUI/Classes/SDUIViewModel.swift
```diff
@@ -0,0 +1,73 @@
...
+    private var onError: ((Error) async -> Void)?
+
+    init(
+        registry: PrimerStepResolverRegistry = .shared,
+        onEvent: @escaping (CodableValue) async throws -> Void,
+        onError: ((Error) async -> Void)? = nil
+    ) {
+        self.onEvent = onEvent
+        self.onError = onError
+        registry.register(self, for: "ui.render")
```
OnurVar, line 35: registry's an actor, this needs await. also nothing's compiling this yet.

henry-cooper-primer: Registry won't be an actor in a further [PR](https://github.com/primer-io/primer-sdk-ios/pull/1844), but I've kept it like this to keep the PRs understandable and manageable. The module is not being depended on at this point so there is no build failure

Modules/PrimerBDCUI/Sources/PrimerUI/Classes/Router.swift
henry-cooper-primer: ~Good point. I'll create a new internal-to-this module nav class~ Removed not needed

OnurVar (approved):

PR #1845, opened 2026-07-28
refactor: Update StepResolver API

henry-cooper-primer:


Modules/PrimerStepResolver/Sources/PrimerStepResolver/Classes/PrimerStepResolver.swift
```diff
@@ -24,7 +24,7 @@ public struct StepResolutionResult: Sendable {
 }
 
 @_spi(PrimerInternal)
-public actor PrimerStepResolverRegistry {
+public final class PrimerStepResolverRegistry {
```
OnurVar, line 27: Is `resolvers` still safe without the actor? `register` writes on whatever thread the caller is on, and `resolve` reads on a background one.

Modules/PrimerBDCCore/Sources/PrimerBDCCore/Classes/PrimerStepOrchestrator.swift
```diff
@@ -44,6 +44,7 @@ final class PrimerStepOrchestrator: StepOrchestrating {
 
     func start(rawSchema: String, initialState: CodableValue) async throws {
         await registry.register(harness, for: "url.open")
+        registry.register(harness, for: "url.open")
```
OnurVar, line 47: is the `await` line leftover?

OnurVar (commented):
Rename itself looks clean. Main question is what serializes the registry now that it isn't an actor.

Modules/PrimerStepResolver/Sources/PrimerStepResolver/Classes/PrimerStepResolver.swift
henry-cooper-primer: I'm keen to avoid awaiting on registering because it creates other race-condition probablites where the resolution fires before the registration completes. I've added an NSLock to prevent concurrent hits to the dictionary

OnurVar (approved):

PR #1846, opened 2026-07-29
feat: Complete BDCUI

henry-cooper-primer:
# Description

Note I have not added the module dependency from PrimerSDK to PrimerBDCUI (either via SPM or Pods) to save on release time


Modules/PrimerBDCUI/Sources/PrimerUI/Model/UIDefinition/Component.swift
```diff
@@ -0,0 +1,90 @@
...
+        case .text: self = .text(try container.decodeProps(TextProps.self))
+        case .textField: self = .textField(try container.decodeProps(TextFieldProps.self))
+        case .spacer: self = .spacer
+        }
+    }
+	
+}
+
+private extension KeyedDecodingContainer<Component.CodingKeys> {
+    func decodeProps<T>(_ type: T.Type) throws -> T where T: Decodable { try decode(T.self, forKey: .props) }
```
OnurVar, line 87: What happens if a component arrives without `props`? This throws, and it takes down the whole tree rather than just that one component.

Modules/PrimerBDCUI/Sources/PrimerUI/Views/Elements/ComponentView.swift
```diff
@@ -0,0 +1,192 @@
...
+        slots: Slots?,
+        content: @escaping (() -> C)
+    ) {
+        componentArgs = UIComponentArgs(fieldID: fieldID, component: component)
+        self.content = content
+        self.slots = slots
+        self.focusedFieldID = focusedFieldID
+        self.visible = switch visible {
+        case let .bool(bool): bool
+        case .string: false
```
OnurVar, line 41: Does the engine ever send `visible` as a string? If it does, the component stays hidden for good.

Modules/PrimerBDCUI/Sources/PrimerUI/Model/UIDefinition/UIElement.swift
```diff
@@ -0,0 +1,69 @@
+//
+//  UIElement.swift
+//
+//  Copyright © 2026 Primer API Ltd. All rights reserved. 
+//  Licensed under the MIT License. See LICENSE file in the project root for full license information.
+
+enum Element: UI {
```
OnurVar: Is `Element` used anywhere? I couldn't find a reference to it.

OnurVar (commented):
A few questions on the decoding side, mostly about what the engine guarantees to send.

Modules/PrimerBDCUI/Sources/PrimerUI/Views/Elements/ComponentView.swift
henry-cooper-primer: This is needed prior to the state processor processing the CEL into a legit bool. Visible arrives as a string and then is quickly processed into a bool

Modules/PrimerBDCUI/Sources/PrimerUI/Model/UIDefinition/UIElement.swift
henry-cooper-primer: removed cheers

Modules/PrimerBDCUI/Sources/PrimerUI/Model/UIDefinition/Component.swift
henry-cooper-primer: At the moment, it's guaranteed all items will have props via backend contract

OnurVar (approved):

PR #1852, opened 2026-08-01
fix: Move SDK file cache into Library/Caches/primer

OnurVar:
# Description

Stacked on #1851 — merge that one first. This PR's own diff is the last commit (4 files); it sits on #1851 because `AnalyticsStorageTests` is flaky on master (it uses the production analytics file) and #1851 is what fixes it.

The SDK caches downloaded payment-method images loose in the host app's documents directory. Three problems:

- **The base64 write path is broken.** `File.init(base64Data:)` writes to a `primer/` subdirectory of documents that is never created, so the write fails silently, while `localUrl`/`data` read from the documents root. Icons delivered inline as base64 (via `ImageFileProcessor`) therefore never cache. The download path was unaffected — it writes to `localUrl` and reads back correctly.
- **Filename collisions with the host app.** Cache filenames are unscoped in a directory the host app also owns, and `Downloader` returns early when a file already exists — so a host-app file of the same name can be served as a cached SDK image.
- **`ImageManager.clean()` deletes every `.png` in the documents directory**, including files the SDK does not own. Latent today: `ImageManager` is internal and nothing calls `clean()` outside tests — but it is a loaded gun.

Documents is also iCloud-backed, and re-downloadable icons shouldn't be.

Changes:

- `File.localUrl` now points into `Library/Caches/primer/`
- the base64 write targets `localUrl`, so it lands where reads look
- `File` and `Downloader` create the directory before writing
- `ImageManager.clean()` sweeps only `Caches/primer/`; the test now proves a host-app `.png` in documents survives it
- the cached-image test is hermetic and asserts the cache hit (it previously swallowed all errors, asserted nothing, and attempted a real download of example.com on every run)

Migration: files cached at the old location by earlier versions are orphaned and re-downloaded once into the new location. The old location is deliberately not swept — scanning the host app's documents directory is the behavior being removed here.

No breaking changes (`ImageManager` and `File` caching are internal / SPI).

Verification: `ImageManagerTests` ×3 in isolation, full suite ×2 (923 passed, 0 failures, 0 crashes each).

# Contributor Checklist

# Reviewer Checklist

# Before Merging

# Other Stuff

### DIFF
--- a/Modules/PrimerCore/Sources/PrimerCore/Models/File.swift
+++ b/Modules/PrimerCore/Sources/PrimerCore/Models/File.swift
@@ -18,18 +18,24 @@ public typealias FileExtension = String
 
 @_spi(PrimerInternal) open class File: LogReporter {
 
+    /// Directory holding all SDK-cached files, namespaced away from the host app's data.
+    /// Lives under Caches: the contents are re-downloadable and should not be backed up.
+    public static var cacheDirectoryUrl: URL? {
+        FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first?
+            .appendingPathComponent("primer", isDirectory: true)
+    }
+
     public var fileName: FileName
     public var fileExtension: FileExtension?
     public var localUrl: URL? {
-        guard let documentDirectoryUrl = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask).first else { return nil }
+        guard let cacheDirectoryUrl = Self.cacheDirectoryUrl else { return nil }
 
         var tmpFilename: String = self.fileName
         if let fileExtension = self.fileExtension {
             tmpFilename += "." + fileExtension
         }
 
-        let fileLocalUrl = documentDirectoryUrl.appendingPathComponent(tmpFilename)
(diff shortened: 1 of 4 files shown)


Modules/PrimerCore/Sources/PrimerCore/Models/File.swift
```diff
@@ -18,18 +18,29 @@ public typealias FileExtension = String
...
 @_spi(PrimerInternal) open class File: LogReporter {
 
+    /// Directory holding all SDK-cached files, namespaced away from the host app's data.
+    /// Lives under Caches: the contents are re-downloadable and should not be backed up.
+    public static var cacheDirectoryUrl: URL? {
+        FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask).first?
+            .appendingPathComponent("primer", isDirectory: true)
+    }
+
+    public static func ensureCacheDirectoryExists() {
```
henry-cooper-primer: nit: with other funcs

henry-cooper-primer (dismissed):

henry-cooper-primer (approved):

PR #1856, opened 2026-08-07
fix: Allow metadata textfield to accept raw json

henry-cooper-primer:


Debug App/Sources/View Controllers/MerchantSessionAndSettingsViewController.swift
```diff
@@ -705,13 +705,17 @@ class MerchantSessionAndSettingsViewController: UIViewController {
...
-            metadataTextField.text?.components(separatedBy: ",").forEach {
-                let tuple = String($0).components(separatedBy: "=")
-                guard tuple.count == 2
-                else { return }
-                let key = tuple[0].trimmingCharacters(in: .whitespaces)
-                let value = tuple[1].trimmingCharacters(in: .whitespaces)
-                try? metadataDict.add(.string(value), forKey: key)
+            if case let .dictionary(json)? = Metadata(jsonObject: metadata) {
+                json.forEach { try? metadataDict.add($0.value, forKey: $0.key) }
+            } else {
```
OnurVar, line 710: If the pasted JSON has a typo, this falls through to the `key=value` parser, which silently adds nothing — should that surface as an error instead?

Debug App/Sources/Model/Metadata.swift
```diff
@@ -1,22 +1,44 @@
...
 
     init(from decoder: Decoder) throws {
         let container = try decoder.singleValueContainer()
-        if let string = try? container.decode(String.self) {
-            self = .string(string)
+        if container.decodeNil() {
+            self = .null
         } else if let bool = try? container.decode(Bool.self) {
             self = .bool(bool)
+        } else if let int = try? container.decode(Int.self) {
```
OnurVar, line 34: `Int` before `Double` means a pasted `1.0` re-encodes as `1` — was that intended, or should `Double` be tried first?

OnurVar (commented):
Two things on the raw-JSON path — both about the pasted text not making it out intact.

OnurVar (approved):

PR #1858, opened 2026-08-13
fix(components): use design tokens for colours that ignored the merchant theme (ORC-8114)

OnurVar:
# Description

ORC-8114

Two things in Checkout Components.

The card-network border on co-badged cards stayed light in dark mode.

And 20 call sites looked like they followed the merchant's theme but did not: button labels, the success tick, spinners, several icons. A merchant who set those colours saw nothing change. They read their tokens now.

**A merchant who sets nothing.** About a dozen resolved colours move, all of them towards the design tokens. The biggest group is every hardcoded `.blue`, which goes from the system blue `#007AFF` to brand `#2f98ff`, or to `text.link` on the country screen's Cancel. The Apple Pay and QR spinners go from the system grey to brand. The QR border goes from `Color.gray.opacity(0.5)` to the border token `#e0e0e0`. The `.primary`, `.secondary` and `.green` sites take their token values. Two sites swap one token for a more accurate one: the co-badge selector border moves to `border.outlined.selected`, and the dynamic-form error border from `text.negative` to `border.outlined.error`. Every swap was checked against the light and dark token files first.

**A merchant who has themed the checkout.** Those 20 sites respond to their colours now. Nothing they can already set changes meaning or resolves differently.

**Public API.** No change. No token is added, renamed or removed, and `ColorOverrides` is untouched.

**Not in scope.** The token values themselves, the width tokens, the input-background split and the `primerColorBackground` rename. Those are the later PRs in this stack.

# Other Notes

`primerColorGray700` is a dead public override once `CheckoutColors.gray700` goes with this PR. It is removed in #1864, not here.

# Manual Testing

Not run on a device or simulator.

# Screenshots

None.

# Contributor Checklist

- [ ]  All status checks have passed prior to code review
- [ ]  I have added unit tests to a reasonable level of coverage where suitable
- [ ]  I have added UI tests to new user flows, if applicable
- [ ]  I have manually tested newly added UX
- [ ]  I have open a documentation PR, if applicable

# Reviewer Checklist

- [ ]  I have verified that a suitable set of automated tests has been added
- [ ]  I have verified that the title prefix aligns to the code changes + whether a release is expected after merging the PR
- [ ]  I have verified the documentation PR aligns with this PR, if applicable

# Before Merging

- [ ]  If introducing a breaking change, I have communicated it internally
- [ ]  Any related documentation PRs are ready to merge

# Other Stuff

[...]

PR #1859, opened 2026-08-14
feat(components): let merchants colour input fields separately from the sheet (ORC-8118)

OnurVar:
# Description

ORC-8118

A merchant asked for this in GitHub discussion #1832: one colour setting painted both the input fields and the whole sheet, so they could not colour the fields without tinting the checkout.

The cause was in the token file, where `background` was both a colour and a group of colours. Style Dictionary built that fine. The runtime Swift flattener is what broke: it collapsed `background` to a single leaf and dropped the 13 tokens underneath it in silence, including the one for input fields. Reshaping the file into `background.primary` and `background.secondary`, which the web token file already has, gives the input-field colour back.

Input fields read their own colour now, and their text does too.

**A merchant who sets nothing.** One default moves. The field fill was `gray.100`; it now reads `background.outlined.default`, which aliases the sheet colour the way Figma and web do. A field goes `#f5f5f5` to `#ffffff` in light, and `#292929` to `#171619` in dark. In dark it becomes the sheet colour, with the border doing the separating. That matches design. Before this, iOS was the one platform tinting it. Everything else resolves to what it did before.

**A merchant who has themed the checkout.** They can set the field fill and the field text on their own now. If they set only the sheet colour, the fields follow it, so their fields do not turn white. Tests cover that.

**Public API, breaking, no alias.** `ColorOverrides.primerColorBackground` becomes `primerColorBackgroundPrimary`. Anyone on the 3.0 beta who sets it stops compiling, with no deprecated alias to soften it. Android renames the same property in primer-io/primer-sdk-android#1313, so both natives break in the same batch and merchants migrate once. The docs PR and the breaking change note are still to do.

Fifteen tokens are added to `ColorOverrides`: `primerColorBackgroundSecondary`, the seven `primerColorBackgroundOutlined*`, the six `primerColorBackgroundTransparent*`, and `primerColorTextOutlinedDefault`. All are additive. Two of the hover ones come straight back out in #1870, since hover cannot fire on touch.

**Deliberate divergence from Figma.** `text.outlined.default` is not in Figma and was not on web upstream; I added it there. Keeping it is a decision, taken on the review thread rather than by default. Once a merchant can set a dark field fill on a light sheet, the typed text still follows `textPrimary`, stays dark and becomes unreadable with no way out. This token is the way out. Raised with design so Figma catches up, since a token that lives only in our code can be wiped the next time someone regenerates.

**Not in scope.** Token values, width tokens and the button recipe. Those are #1864 and #1870.

# Other Notes

Stacked on #1858. Review that first.

Also here: one Swift fallback colour disagreed with the token file, the country search field moves to `background.secondary` (same resolved grey, named for what it is), a duplicate colour accessor is gone with its one caller moving to `textNegative`, and the billing-address error border moves from `text.negative` to `border.outlined.error` so it matches the card form.

From review: palette overrides are injected before references resolve, so overriding `gray300` now cascades into `border.outlined.default` instead of only changing the leaf. A merchant colour whose RGB components cannot be read is skipped and logged rather than dropped in silence. The dead `resolveReferences` pass with the inverted ternary is deleted. New test flattens the shipped `base.json` the way production does and asserts every colour leaf has a `DesignTokens` property to land in; it is what found the dead `primerColorGray700` override.

# Manual Testing

Not run on a device or simulator.

# Screenshots

None attached. The one visible change is the field fill described above.

# Contributor Checklist

# Reviewer Checklist

# Before Merging

# Other Stuff

### DIFF
--- a/Sources/PrimerSDK/Classes/CheckoutComponents/Scope/PrimerCheckoutTheme.swift
+++ b/Sources/PrimerSDK/Classes/CheckoutComponents/Scope/PrimerCheckoutTheme.swift
@@ -94,6 +94,19 @@ public struct ColorOverrides: Equatable {
   public let primerColorBackgroundPrimary: Color?
   public let primerColorBackgroundOutlinedDefault: Color?
   public let primerColorTextOutlinedDefault: Color?
+  public let primerColorBackgroundSecondary: Color?
+  public let primerColorBackgroundOutlinedHover: Color?
+  public let primerColorBackgroundOutlinedActive: Color?
+  public let primerColorBackgroundOutlinedDisabled: Color?
+  public let primerColorBackgroundOutlinedLoading: Color?
+  public let primerColorBackgroundOutlinedSelected: Color?
+  public let primerColorBackgroundOutlinedError: Color?
+  public let primerColorBackgroundTransparentDefault: Color?
+  public let primerColorBackgroundTransparentHover: Color?
+  public let primerColorBackgroundTransparentActive: Color?
+  public let primerColorBackgroundTransparentDisabled: Color?
+  public let primerColorBackgroundTransparentLoading: Color?
+  public let primerColorBackgroundTransparentSelected: Color?
   public let primerColorTextPrimary: Color?
   public let primerColorTextSecondary: Color?
   public let primerColorTextPlaceholder: Color?
@@ -153,6 +166,19 @@ public struct ColorOverrides: Equatable {
     primerColorBackgroundPrimary: Color? = nil,
     primerColorBackgroundOutlinedDefault: Color? = nil,
     primerColorTextOutlinedDefault: Color? = nil,
+    primerColorBackgroundSecondary: Color? = nil,
+    primerColorBackgroundOutlinedHover: Color? = nil,
+    primerColorBackgroundOutlinedActive: Color? = nil,
+    primerColorBackgroundOutlinedDisabled: Color? = nil,
+    primerColorBackgroundOutlinedLoading: Color? = nil,
+    primerColorBackgroundOutlinedSelected: Color? = nil,
+    primerColorBackgroundOutlinedError: Color? = nil,
+    primerColorBackgroundTransparentDefault: Color? = nil,
+    primerColorBackgroundTransparentHover: Color? = nil,
+    primerColorBackgroundTransparentActive: Color? = nil,
+    primerColorBackgroundTransparentDisabled: Color? = nil,
+    primerColorBackgroundTransparentLoading: Color? = nil,
+    primerColorBackgroundTransparentSelected: Color? = nil,
     primerColorTextPrimary: Color? = nil,
     primerColorTextSecondary: Color? = nil,
     primerColorTextPlaceholder: Color? = nil,
@@ -199,6 +225,19 @@ public struct ColorOverrides: Equatable {
     self.primerColorBackgroundPrimary = primerColorBackgroundPrimary
     self.primerColorBackgroundOutlinedDefault = primerColorBackgroundOutlinedDefault
     self.primerColorTextOutlinedDefault = primerColorTextOutlinedDefault
+    self.primerColorBackgroundSecondary = primerColorBackgroundSecondary
+    self.primerColorBackgroundOutlinedHover = primerColorBackgroundOutlinedHover
+    self.primerColorBackgroundOutlinedActive = primerColorBackgroundOutlinedActive
+    self.primerColorBackgroundOutlinedDisabled = primerColorBackgroundOutlinedDisabled
+    self.primerColorBackgroundOutlinedLoading = primerColorBackgroundOutlinedLoading
+    self.primerColorBackgroundOutlinedSelected = primerColorBackgroundOutlinedSelected
+    self.primerColorBackgroundOutlinedError = primerColorBackgroundOutlinedError
+    self.primerColorBackgroundTransparentDefault = primerColorBackgroundTransparentDefault
+    self.primerColorBackgroundTransparentHover = primerColorBackgroundTransparentHover
+    self.primerColorBackgroundTransparentActive = primerColorBackgroundTransparentActive
+    self.primerColorBackgroundTransparentDisabled = primerColorBackgroundTransparentDisabled
+    self.primerColorBackgroundTransparentLoading = primerColorBackgroundTransparentLoading
+    self.primerColorBackgroundTransparentSelected = primerColorBackgroundTransparentSelected
     self.primerColorTextPrimary = primerColorTextPrimary
     self.primerColorTextSecondary = primerColorTextSecondary
     self.primerColorTextPlaceholder = primerColorTextPlaceholder
--- a/Sources/PrimerSDK/Classes/CheckoutComponents/Scope/PrimerCheckoutTheme.swift
+++ b/Sources/PrimerSDK/Classes/CheckoutComponents/Scope/PrimerCheckoutTheme.swift
@@ -92,9 +92,8 @@ public struct ColorOverrides: Equatable {
   // MARK: Semantic UI Colors (matching internal DesignTokens)
 
   public let primerColorBackgroundPrimary: Color?
-  public let primerColorBackgroundOutlinedDefault: Color?
-  public let primerColorTextOutlinedDefault: Color?
   public let primerColorBackgroundSecondary: Color?
+  public let primerColorBackgroundOutlinedDefault: Color?
   public let primerColorBackgroundOutlinedHover: Color?
   public let primerColorBackgroundOutlinedActive: Color?
   public let primerColorBackgroundOutlinedDisabled: Color?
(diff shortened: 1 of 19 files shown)


OnurVar:
Pushed b38bd1bc3: iOS has two copies of the token file and I had only fixed one, so the next regeneration would have undone this PR.

Worth knowing separately: `DesignTokens.swift` says it is generated, but it has been hand-edited since, and regenerating it now produces something that will not compile. Needs a header warning or the generator brought back in line.

Sources/PrimerSDK/Classes/CheckoutComponents/Scope/PrimerCheckoutTheme.swift
```diff
@@ -91,7 +91,22 @@ public struct ColorOverrides: Equatable {
 
   // MARK: Semantic UI Colors (matching internal DesignTokens)
 
-  public let primerColorBackground: Color?
+  public let primerColorBackgroundPrimary: Color?
```
bthebladeprimer, line 94: Decision one. Do we rename this while Android does not?

Android still calls it `primerColorBackground`, at `LightColorTokens.kt:60`, and Android is our source of truth for the CheckoutComponents public API. The rename also breaks source for merchants on the 3.0 beta. Nine files in `public-docs` use the old name, including the versioned iOS reference at `sdk/ios-checkout/v3.0.0-beta/api/theming/color-tokens.mdx` on lines 66 and 130.

My proposal is to follow Android on the public name and keep your token file reshape. The two are separable. The JSON can keep `background.primary` to match web, and `ColorOverrides` can keep `primerColorBackground` mapped onto it. Then the merchant fix ships now, and we rename on both platforms together when Android does.

If we rename now instead, we need the Android decision, a docs PR and a breaking change note first. Can we ask Semir and Niall to weigh in?

bthebladeprimer (commented):
I do not block this. Two cross platform decisions are on the file, and we should make them as a team.

Two fixes for the description. The `background.primary` and `background.secondary` reshape is not new, because the web token file already holds both. It is at `primer-sdk/packages/design-tokens/tokens/light.json`. Also "the build stopped there" is not what happened. The Style Dictionary build was fine, and the runtime Swift flattener dropped the subtree in silence.

Four smaller items, none of them blocking:

- The fallback for `CheckoutColors.inputBackground` is `Color(.systemGray6)`, but the token resolves to `#ffffff`. Please align the two.
- Palette overrides get frozen to RGBA through `UIColor(...).getRed`. A merchant color where `getRed` returns false skips the cascade in silence, while the leaf still changes.
- `resolveReferences` at `DesignTokensProcessor.swift:39` has an inverted ternary. It drops the output of any pass where everything resolved. That is harmless today, and cheap to fix while you are in the file.
- Please add a test that every leaf in `base.json` has a matching `DesignTokens` property. It is about 20 lines. I can write it if you prefer.

Sources/PrimerSDK/Classes/CheckoutComponents/Scope/PrimerCheckoutTheme.swift
```diff
@@ -91,7 +91,22 @@ public struct ColorOverrides: Equatable {
...
+  public let primerColorBackgroundOutlinedDefault: Color?
+  public let primerColorTextOutlinedDefault: Color?
```
bthebladeprimer: Decision two. This is the one token we would own alone.

I compared our `base.json` leaf by leaf against the web `light.json` and the Android `base.json`. Everything else in this PR matches web. `primer.color.text.outlined.default` is the only exception, and this PR makes it public merchant API.

My proposal is to follow web and Android again. The merchant asked for the input background in discussion #1832, and `background.outlined.default` covers that and exists on web. So drop `text.outlined.default`, point `CheckoutColors.inputText` back at `primerColorTextPrimary`, and ship the background half now. If we still want an input text token, we add it through design, then web, then Android and iOS.

One small thing in the same file. `primerColorTextOutlinedDefault` at line 168 comes before `primerColorTextPrimary` at line 181, and Swift enforces the argument order. So `ColorOverrides(primerColorTextPrimary:, primerColorTextOutlinedDefault:)` does not compile. Please move one of them.

Tests/Primer/CheckoutComponents/Tokens/DesignTokensManagerTests.swift
```diff
@@ -739,6 +739,128 @@ final class DesignTokensManagerTests: XCTestCase {
...
+        sut.applyTheme(PrimerCheckoutTheme(colors: ColorOverrides(primerColorGray300: .pink)))
+
+        // When
+        try await sut.fetchTokens(for: .light)
+
+        // Then the border has moved, because it aliases the overridden palette entry.
+        // It is not compared against the passed-in Color: palette overrides resolve through the
+        // token JSON, so the border holds a hex-derived Color while gray300 keeps the raw one.
+        let themed = try XCTUnwrap(sut.tokens)
+        XCTAssertNotEqual(themed.primerColorBorderOutlinedDefault, unthemedBorder)
```
bthebladeprimer, line 758: This assertion and its dark mode twin at line 808 check only that the color moved. A wrong value that still moved would pass.

`border.outlined.default` aliases `gray.300`. So override `gray300` with a known color, then compare against the hex derived `Color` for that value. Please add one assertion on the resolved value.

Sources/PrimerSDK/Classes/CheckoutComponents/Scope/PrimerCheckoutTheme.swift
OnurVar: Premise is off, my fault for not linking the other PRs. Android renames the same property in primer-io/primer-sdk-android#1313, `primerColorBackground` becomes `primerColorBackgroundPrimary`, and web has been on `background.primary` upstream for a while. So iOS is catching up rather than diverging.

Which makes your point bigger, not smaller. If both natives rename then two public APIs break for anyone on the 3.0 beta. Doing the docs PR and the breaking change note, covering iOS and Android together.

Sources/PrimerSDK/Classes/CheckoutComponents/Scope/PrimerCheckoutTheme.swift
OnurVar: All four platforms add this token in this batch. Android has `primerColorTextOutlinedDefault` in #1313, RN has `textOutlinedDefault` in #412, web has it in !2205.

But you are right about the part that matters. It is not on web upstream because I added it there myself, so design never signed off. Taking it to them. If they say no we drop it from all four together rather than from iOS alone.

Argument order is a real bug, thanks. Reordered so base tokens come before their outlined variants, in the properties, the init and the assignments.

Tests/Primer/CheckoutComponents/Tokens/DesignTokensManagerTests.swift
OnurVar: Agreed, asserting it moved is worth little. It now overrides `gray300` with components that survive the getRed round trip exactly and compares the resolved value.

And yes please on the base.json coverage test, I have added it. It flattens the shipped file the way production does and asserts every colour leaf has a `DesignTokens` property to land in. It caught `primerColorGray700` as the one property with no token behind it, which is the dead override you spotted on #1858.

OnurVar:
Picked up the four from the review body.

- `inputBackground` fallback now matches its token, `.white` rather than `systemGray6`.
- A palette override we cannot read now logs instead of vanishing.
- On `resolveReferences`, it turned out to be more than an inverted ternary. I ran it against the shipped `base.json` and `base.json` merged with `dark.json` and its output is byte identical to its input in both cases, for all 62 aliased tokens. Nothing was resolving there. `resolveFlattenedReferences` does the real work and has no such bug, so I deleted the function and its call site rather than fix the ternary, which keeps behaviour identical. Its four tests only asserted the input passed through unchanged, so they went too; I moved the two cases the flattened suite did not already cover onto it.
- Coverage test added, see the thread above.

Sources/PrimerSDK/Classes/CheckoutComponents/Scope/PrimerCheckoutTheme.swift
OnurVar: Following up, and correcting myself.

I checked Figma directly rather than waiting. `color/text/*` there is primary, secondary, placeholder, disabled, negative and link. There is no `text/outlined/default`. `background/outlined/default` and the whole `border/outlined/*` family are there, so the absence looks deliberate rather than an oversight.

So you were right, and my earlier answer was circular. It is on all four platforms only because I put it on all four.

We are keeping it anyway, which contradicts what I said above about dropping it if design said no, so flagging that plainly rather than letting it slide. The reason is that ORC-8118 makes the gap reachable: once a merchant can set the input fill on its own, they can set a dark fill on a light sheet, and the typed text still follows `textPrimary` and stays dark. Unreadable, with no way out. The token is the way out.

Raising it with design so Figma catches up. Not for permission, but because a token that lives only in our code can get wiped the next time someone regenerates from Figma. That nearly happened on Android in this batch, where the colour templates were still emitting the pre-rename name.

bthebladeprimer (approved):

OnurVar:

PR #1863, opened 2026-08-20
chore: Ingest current attempt in BDC flow

henry-cooper-primer:


Modules/PrimerBDCEngine/Sources/Extensions/PrimerBDCEngine+Scripts.swift
```diff
@@ -79,7 +79,7 @@ extension PrimerBDCEngine {
         (async () => {
             try {
                 const processor = await StateProcessor.createStateProcessor(__schema, JSON.parse(__context));
-                const result = await processor.applyResult(JSON.parse(__state), __actionId, __outcome, __data);
+                const result = await processor.applyResult(JSON.parse(__state), __actionId, __outcome, JSON.parse(__data));
```
OnurVar, line 82: `data` is `String?` and the Safari harness resolves with nil, so `JSON.parse(undefined)` throws. Worth a guard?

OnurVar: `{}` where it used to be `undefined`. Intentional?

henry-cooper-primer: yes intentional, the state processor wants an empty map where there is nil data

OnurVar (dismissed):

OnurVar (approved):

PR #1864, opened 2026-08-21
fix(components): button labels in dark mode, input border width, dead dark token class (ORC-8179)

OnurVar:
# Description

ORC-8179, ORC-8178 and ORC-8198.

Three follow-ups from checking our theming against Figma.

**Pay button labels were invisible in dark mode (ORC-8179).** The primary button fills with the text colour, which flips from near-black to near-white between modes, but the label was hardcoded white. Seven buttons across six call sites showed white on near-white, at 1.15:1. The label follows the fill now, and the disabled state follows too. The three buttons that really sit on brand blue, Apple Pay, the error screen retry and the dynamic-form submit, keep a fixed white.

**Only the vaulted CVV field thickened its border (ORC-8178).** Figma asks for 2pt on focus and error; the card form and the two dynamic-form renderers drew 1pt in every state. All four renderers follow Figma now, with the resting border unchanged at 1pt. Billing-address fields track errors but not focus, so those thicken on error only.

**Deleted a dark-mode colour file nothing used (ORC-8198).** Dark mode is resolved at runtime from the token files, so the generated `DesignTokensDark` class was dead. Removed it, its Style Dictionary build step, the workspace entry and the stale docs.

**A merchant who sets nothing.** Light mode is not unchanged, contrary to what this description said before.

- Any focused or errored input field draws a 2pt border instead of 1pt, in both modes. The field's outer size does not move.
- A disabled Pay button changes fill from `gray300` `#e0e0e0` to `background.outlined.disabled` `#f5f5f5`, so it gets lighter, and its label changes from white to `textDisabled` `#bdbdbd`. Same pairing web and Android use. In light mode the old white-on-grey was 1.32:1 and unreadable, so this is a fix rather than a regression, but it is a visible change.
- Dark mode: the enabled Pay button label goes from 1.15:1 to 15.67:1, and the disabled label from 2.49:1 to a readable pairing.

**A merchant who has themed the checkout.** The disabled button now reads `primerColorBackgroundOutlinedDisabled` and `primerColorTextDisabled` instead of `primerColorGray300` and a literal white, so both states are themeable. Focus and error widths are still hardcoded here; #1870 moves them onto tokens.

**Public API, breaking.** `ColorOverrides.primerColorGray700` is removed, with no alias. It was dead once `CheckoutColors.gray700` went in #1858: iOS exposed a ninth grey the other SDKs do not have, and nothing behind it in `base.json`.

**Not in scope.** The error border colour matches Figma exactly, so its low contrast is a design decision rather than our bug, and it is noted on ORC-8178. White on brand `#2f98ff` is 2.98:1, just under the 3:1 floor; raised with design alongside the error border rather than fixed here.

# Other Notes

Stacked on #1859. Review that first.

# Manual Testing

Not run on a device or simulator. Worth checking a disabled Pay button and a field in error, in both modes.

# Screenshots

None.

# Contributor Checklist

# Reviewer Checklist

# Before Merging

# Other Stuff


Sources/PrimerSDK/Classes/CheckoutComponents/Internal/UI/Design/CheckoutColors.swift
```diff
@@ -80,7 +80,14 @@ enum CheckoutColors {
     tokens?.primerColorIconPositive ?? Color(.systemGreen)
   }
 
-  static func white(tokens _: DesignTokens?) -> Color { .white }
+  /// Label/spinner colour on the primary button, which fills with `textPrimary` — so this has to
+  /// invert with it or the label disappears in dark mode.
+  static func onPrimary(tokens: DesignTokens?) -> Color {
```
bthebladeprimer: `onPrimary` follows `primerColorBackgroundPrimary`, but these buttons do not always fill with `textPrimary`. The disabled fill is `gray300`. So the inversion is right for the enabled state and wrong for the disabled state.

In dark mode the disabled label goes from 7.23:1 to 2.49:1, on fill `#575757` with label `#171619`. In light mode it is 1.32:1 before and after, so that one is older than your PR.

Four buttons and their spinner tints use this accessor. They are `CardFormScreen.swift:124`, `BillingAddressRedirectScreen.swift:260`, `PaymentMethodsDefaults.swift:223` and `AchUserDetailsView.swift:92`.

Please pick one of two fixes. Derive the label color from the button state, so the disabled fill gets `textDisabled`. Or copy Android, which uses `primerColorGray000` in light mode and `primerColorGray900` in dark mode, at `PrimerTheme.kt:58`. I prefer the Android one. If you agree, fix the light disabled state in the same pass.

Sources/PrimerSDK/Classes/CheckoutComponents/Internal/UI/Design/CheckoutColors.swift
```diff
@@ -80,7 +80,14 @@ enum CheckoutColors {
...
+    tokens?.primerColorBackgroundPrimary ?? .white
+  }
+
+  /// Label/spinner colour on a brand-filled surface. Brand is a fixed blue in both modes.
+  static func onBrand(tokens _: DesignTokens?) -> Color { .white }
```
bthebladeprimer: `ErrorScreen.swift:76` still uses a literal `.white` on a `buttonPrimary` fill. Please route it through this accessor.

White on brand `#2f98ff` is 2.98:1, just under the 3:1 floor. Please add that to the error border question you already raised with design.

Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Presentation/Components/Infrastructure/Containers/PrimerInputFieldContainer+Styling.swift
```diff
@@ -48,7 +48,12 @@ extension PrimerInputFieldContainer {
 @available(iOS 15.0, *)
 extension PrimerInputFieldContainer {
   var fieldCornerRadius: CGFloat { PrimerRadius.small(tokens: tokens) }
-  var textFieldContainerBackgroundLineWidth: CGFloat { PrimerBorderWidth.standard(tokens: tokens) }
+  // Design specifies a 2pt stroke for the focus and error states, 1pt otherwise.
+  var textFieldContainerBackgroundLineWidth: CGFloat {
```
bthebladeprimer, line 52: This reaches only one of the four input renderers. Three others keep a fixed 1pt on error and on focus:

- `BillingAddressRedirectScreen.swift:176`
- `BillingAddressRedirectScreen.swift:210`
- `FormRedirectScreen.swift:229`

So an errored card form field gets a 2pt border, and an errored billing address field keeps 1pt.

The shared web token file already models the state widths. It holds `primer.width.default` 1, `primer.width.focus` 2, `primer.width.error` 2 and `primer.width.selected` 2, at `primer-sdk/packages/design-tokens/tokens/light.json`. If you add `primer.width.*` to our token file, all four renderers can read the width from one place.

bthebladeprimer (changes requested):
The dark mode fix is right. The enabled Pay button goes from 1.15:1 to 15.67:1.

One state is missed, and that is why I request changes. `CheckoutColors.onPrimary` follows the sheet token, but the button fill changes with the state. The disabled label drops to 2.49:1 in dark mode. I left the numbers and two options on the accessor.

Sources/PrimerSDK/Classes/CheckoutComponents/Internal/UI/Design/CheckoutColors.swift
OnurVar: Taking option one, the label now derives from the button state.

Two notes. `PrimerTheme.kt:58` is Android's rule for brand filled surfaces rather than disabled ones. What Android's button actually does is fall back to `primerColorTextDisabled`, and web does the same with `--primer-color-text-disabled`, so that is what I followed.

The label alone was not enough though. `textDisabled` on the old `gray300` fill lands around 1.4:1, no better than the white it replaces. Web pairs `text-disabled` with `background-outlined-disabled`, so `buttonDisabled` now returns `primerColorBackgroundOutlinedDisabled` and the two move together. Light disabled is fixed in the same pass.

One you did not list: `FormRedirectScreen` had the same pairing through `onBrand` and `buttonDisabled`, so it is covered too.

Sources/PrimerSDK/Classes/CheckoutComponents/Internal/UI/Design/CheckoutColors.swift
OnurVar: Fixed, it goes through `onBrand` now.

Adding the brand on white number to the design question alongside the error border.

Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Presentation/Components/Infrastructure/Containers/PrimerInputFieldContainer+Styling.swift
OnurVar: Confirmed, all three were on a fixed `PrimerBorderWidth.standard`. Fixed. `FormRedirect` tracks focus so it mirrors the shared error-or-focus rule; the two billing address renderers only track errors, so they thicken on error.

On the width tokens I would rather do that separately. Web has `primer.width.*`, but Android is on `thin/medium/thick` and RN on `default/input/strong`, so aligning the vocabulary is a four platform job and does not belong in a button colour PR. Raising a ticket.

OnurVar: Correction: ORC-8228 was deleted. The ticket for this is **ORC-8229** — https://primerapi.atlassian.net/browse/ORC-8229 — and it now covers the whole token alignment rather than just the width names.

The rest of this comment still holds: Android had no width group in its style-dictionary `base.json`, so `thin/medium/thick` were Kotlin defaults rather than generated tokens, and aligning the names there meant writing the generator first.

OnurVar: Raised as ORC-8229: https://primerapi.atlassian.net/browse/ORC-8229

Worth correcting something in my reply above though. I said this was a naming cleanup, but checking the token files properly it is more than that. Web has eight `primer.width.*` tokens and iOS has none in its runtime token JSON at all, so the widths come from hardcoded Swift defaults and a merchant cannot theme them. Android is the same, its `thin/medium/thick` are Kotlin defaults with no entry in style-dictionary. So this is adding a merchant control web already has, not just renaming one.

Also worth flagging while I was in there: Android's colour token templates still emit the pre-rename `primerColorBackground`, so the next token build would have silently undone part of the Android side of this batch.

OnurVar: Retracting two things I said in this thread. An audit pass caught both.

I said a merchant could not theme the widths at all. Wrong — `BorderWidthOverrides` was already public and wired end to end, `PrimerCheckoutTheme.swift:28` through `DesignTokensManager.swift:143` to `PrimerLayout.swift:130`. What was actually broken is narrower: the runtime token JSON had no width group while the build-time copy did, so the two copies disagreed and the values came from Swift defaults. This PR is a rename plus fixing that drift, not a new control.

I also said Android's colour templates would have silently reverted part of #1313 on the next token build. Wrong. All eight occurrences of the old name are inside doc comment blocks; the generator derives property names from `token.name`, so a rebuild would have emitted the new name correctly. The Android commit I described as fixing it changed comment text only.

bthebladeprimer (approved):

OnurVar:

PR #1865, opened 2026-08-21
fix(components): stop the country list covering the sheet colour (ORC-8178)

OnurVar:
# Description

ORC-8178

Two colour bugs on iOS.

**The country list ignored the theme.** It is a SwiftUI `List`, which paints its own background on top of the sheet colour, so a merchant's background never showed on that screen. Rows are transparent now, and on iOS 16 and up the list background is hidden.

**Two selection ticks were hardcoded blue.** The country picker's and Klarna's checkmarks used an accessor that took the theme and threw it away. They read `border.outlined.selected` now, which leaves that accessor unused, so it is removed.

**A merchant who sets nothing.** The two ticks go from the system blue `#007AFF` to brand `#2f98ff`. That trades light for dark: on the white sheet contrast drops from 4.02:1 to 2.98:1, and in dark mode it improves from 4.94:1 to 6.06:1. The brand value is what design specifies, and the sub-3:1 number is on the same question already raised with design about the error border. No pixel moves and no layout changes.

**A merchant who has themed the checkout.** The country screen finally takes their background colour, on iOS 16 and up. Their selected-border colour now paints the two ticks.

**iOS 15.** The fix does not reach it. `scrollContentBackground` is iOS 16 only, and on iOS 15 the rows and the underlying `UITableView` share one system background, so the transparent rows do nothing on their own. The country screen keeps the system background there.

**Public API.** No change.

**Not in scope.** Two items on this ticket turned out not to be bugs, and are noted on the ticket rather than changed: switching the phone to dark mode mid-checkout already reloads the tokens, and the "white flash on open" falls back to a dark colour in dark mode, not white. Two warning triangles keep a hardcoded orange, deliberately: there is no warning colour in the token set at all, the same gap web hit with its BLIK note, so there is nowhere themed to point them.

# Other Notes

Stacked on #1864. Review that first.

The red Unit Tests run was a flake. The same SHA passed on 2026-08-21 in run 32536917307 and failed on the 2026-08-22 re-run in run 32575648070, both with 3022 tests. Re-run since.

# Manual Testing

Not run on a device or simulator. Worth opening the country picker on a dark theme to confirm the background.

# Screenshots

None.

# Contributor Checklist

- [x]  All status checks have passed prior to code review
- [ ]  I have added unit tests to a reasonable level of coverage where suitable
- [ ]  I have added UI tests to new user flows, if applicable
- [ ]  I have manually tested newly added UX
- [ ]  I have open a documentation PR, if applicable

# Reviewer Checklist

[...]

PR #1870, opened 2026-08-28
refactor(components): bring the theme tokens in line with the other SDKs (ORC-8229)

OnurVar:
# Description

[ORC-8229](https://primerapi.atlassian.net/browse/ORC-8229)

Stacked on #1865. Merge that first.

The four SDKs did not agree on their theme tokens. This brings iOS onto the shared set, so a merchant setting a colour, a width or a font finds the same name whichever SDK they are on. It also corrects the token values that had drifted from Figma, which means pixels move.

## What a merchant who sets nothing sees

**Input fields lose 4pt of height.** `size.xxlarge` was 44 and Figma and web both say 40, so the token goes `size.base * 11` to `size.base * 10`. Every field that reads it drops from 44pt to 40pt: the card form fields, the country field, the OTP field, the vaulted CVV field, and the vaulted card icon, which is square and so loses 4pt each way.

**Primary buttons turn brand blue.** They filled with `textPrimary`, which is near-black in light and near-white in dark. They now fill with `brand` `#2f98ff` with a white label, matching Figma and the other SDKs. This is the card form Pay button, the billing address submit, the ACH buttons, Klarna and the vaulted Pay button. A disabled button still greys out.

**A paying button stays blue.** Submitting used to drop the fill to grey while the spinner ran, which read as disabled. The fill and the spinner now stay brand while the request is in flight. Same for the vaulted Pay button, which used to drop to `gray300` with a white spinner on it.

**Co-badge network chips lose their fill.** The selected chip was filled with `gray.100` and the unselected ones with the sheet colour. Both are transparent now, and the selection is shown by the border alone, as in Figma. Tap targets are unchanged.

**The greys become translucent ink over the sheet, not fixed hexes.** Figma and web write them that way; iOS carried opaque hand-picked values. In light mode this is close to a no-op, because each new value resolves within a shade of the old one on a white sheet (`#f5f5f5` to `#2121210a`, `#e0e0e0` to `#21212124`, and so on). Dark mode does move, because the old dark ramp was hand-picked rather than derived:

| token | before | now, resolved over the dark sheet |
| --- | --- | --- |
| `text.secondary` (`gray.600`) | `#c7c7c7` | about `#9d9d9d` |
| `border.outlined.default` (`gray.300`) | `#575757` | about `#363636` |
| `text.disabled` (`gray.400`) | `#858585` | about `#585858` |
| `text.placeholder` (`gray.500`) | `#767577` | about `#767677`, unchanged |

So secondary text and field borders in dark mode sit lower against the sheet than they did. That is what the design says, and it is the same ramp web and Android use, but it is a real contrast change and worth a look on device.

Everything else in here resolves to the same pixels. `loader`, `border.outlined.selected` and `brand` all alias `brand` today, so the spinners, ticks and selected borders that moved onto them do not change colour; they are just separable now. Strokes that were hardcoded `1` now read `width.default`, which is 1. `typography.error` aliases `bodySmall`, and the FormRedirect error text that read `caption` does not change size, because `caption` already mapped to `bodySmall`.

## What a merchant who has themed the checkout sees

Widths become themeable for real. `BorderWidthOverrides` was already public and wired end to end, so a merchant could set them, but the values lived only as Swift defaults and in the build-time token copy. The runtime JSON had no width group at all, so the two copies disagreed and the Swift defaults won. The widths are in the runtime file now, and they take the state names Figma and web use.

Five more spinners read `primerColorLoader` instead of `primerColorBorderOutlinedFocus`, so setting the loader colour now reaches the splash, the payment method list, the country screen, the default loading screen and the form-redirect pending screen. `primerWidthSelected` separates a chosen saved card from a focused field, which shared one width before. `typography.error` separates field error text, which drew at `bodySmall`, so a merchant who wanted smaller error text also shrank every field label.

A merchant who set a grey now tints more of the sheet than before, since the greys composite over the background rather than covering it.

## BREAKING CHANGES

Public API. Breaking for anyone on the 3.0 beta, no aliases:

- `BorderWidthOverrides` becomes `WidthOverrides`.
- `PrimerCheckoutTheme(borderWidth:)` becomes `PrimerCheckoutTheme(width:)`, and the property with it.
- `primerBorderWidthThin`, `Medium` and `Thick` become `primerWidthDefault`, `primerWidthFocus` and `primerWidthError`. `thick` is dropped rather than renamed, because nothing drew with it.
- Four hover colours are removed from `ColorOverrides`: `primerColorBackgroundOutlinedHover`, `primerColorBackgroundTransparentHover`, `primerColorBorderOutlinedHover` and `primerColorBorderTransparentHover`. Hover is a mouse state and cannot fire on touch, so they were settable and inert. They are also gone from both token files.

Added: `WidthOverrides.primerWidthSelected` and `TypographyOverrides.error`.

The four hover colours were public in `3.0.0-beta.1` through `beta.4`. Their removal commit (`47e0effe9`) is already pushed and carries no `BREAKING CHANGE:` footer, so commitizen files it under Refactor and the generated `CHANGELOG.md` will not mark the release breaking. Adding the footer would mean rewriting pushed history, so this section has to be carried into the release notes by hand.

This lands alongside the `primerColorBackground` rename in #1859 and the `primerColorGray700` removal in #1864, so merchants on the beta migrate once rather than three times. The breaking change note and the docs PR cover the whole stack.

## Deliberate divergences and non-goals

The empty `typography.display` group is removed rather than filled.

`text.outlined.default` stays, and it is still not in Figma. That decision and its reasoning are on #1859.

Not in scope: the error border colour and the white-on-brand contrast, both raised with design on ORC-8178. No Drop-in changes; this is Checkout Components only.

# Other Notes

Both token files are updated. iOS keeps a build-time copy under `DesignTokens/` and a runtime copy in `PrimerResources`, and they had drifted before this.

Tests cover the width tokens, the focus and selected widths separately, and that a circular token reference is left unresolved rather than looping.

# Manual Testing

Not device-tested. This one deserves a device pass, because field heights, button fills and the dark grey ramp all move.

# Contributor Checklist

# Reviewer Checklist

# Before Merging

# Other Stuff

[ORC-8229]: https://primerapi.atlassian.net/browse/ORC-8229?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/DesignTokens/tokens/base.json
+++ b/DesignTokens/tokens/base.json
@@ -232,32 +232,32 @@
       "gray": {
         "100": {
           "type": "color",
-          "value": "#f5f5f5ff",
+          "value": "#2121210a",
           "blendMode": "normal"
         },
         "200": {
           "type": "color",
-          "value": "#eeeeeeff",
+          "value": "#21212112",
           "blendMode": "normal"
         },
         "300": {
           "type": "color",
-          "value": "#e0e0e0ff",
+          "value": "#21212124",
           "blendMode": "normal"
         },
         "400": {
           "type": "color",
-          "value": "#bdbdbdff",
+          "value": "#2121214d",
           "blendMode": "normal"
         },
         "500": {
           "type": "color",
-          "value": "#9e9e9eff",
+          "value": "#21212170",
           "blendMode": "normal"
         },
         "600": {
           "type": "color",
-          "value": "#757575ff",
+          "value": "#2121219e",
           "blendMode": "normal"
         },
         "900": {
--- a/DesignTokens/tokens/base.json
+++ b/DesignTokens/tokens/base.json
@@ -313,18 +313,18 @@
         }
       }
     },
-    "borderWidth": {
-      "thin": {
+    "width": {
+      "default": {
         "type": "dimension",
         "value": 1
       },
-      "medium": {
+      "focus": {
         "type": "dimension",
         "value": 2
       },
-      "thick": {
+      "error": {
         "type": "dimension",
-        "value": 3
+        "value": 2
       }
     },
     "radius": {
@@ -350,7 +350,6 @@
       }
     },
     "typography": {
-      "display": null,
       "brand": {
         "type": "string",
         "value": "Inter"
@@ -472,6 +471,28 @@
             "type": "dimension",
             "value": 16
           }
+        }
+      },
+      "error": {
+        "font": {
+          "type": "color",
+          "value": "{primer.typography.body.small.font}"
+        },
+        "weight": {
+          "type": "dimension",
+          "value": "{primer.typography.body.small.weight}"
+        },
+        "size": {
+          "type": "dimension",
+          "value": "{primer.typography.body.small.size}"
+        },
+        "lineHeight": {
+          "type": "dimension",
+          "value": "{primer.typography.body.small.lineHeight}"
+        },
+        "letterSpacing": {
+          "type": "dimension",
+          "value": "{primer.typography.body.small.letterSpacing}"
         }
       }
     },
--- a/DesignTokens/tokens/base.json
+++ b/DesignTokens/tokens/base.json
@@ -553,7 +553,7 @@
       },
       "xxlarge": {
         "type": "dimension",
-        "value": "{primer.size.base} * 11.00"
+        "value": "{primer.size.base} * 10.00"
       },
       "xxxlarge": {
         "type": "dimension",
--- a/DesignTokens/tokens/base.json
+++ b/DesignTokens/tokens/base.json
@@ -33,11 +33,6 @@
             "type": "color",
             "value": "{primer.color.background.outlined.default}"
           },
-          "hover": {
-            "description": "",
-            "type": "color",
-            "value": "{primer.color.background.outlined.default}"
-          },
           "selected": {
             "description": "",
             "type": "color",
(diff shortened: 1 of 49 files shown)


OnurVar:
Checked the batch against Figma directly rather than reasoning from the token files. Three things worth knowing, two of which back this PR and one of which is a bug none of these PRs fix.

**Confirms this PR.** Figma's width tokens are `default 1`, `focus 2`, `selected 2`, `error 2`, exactly the names and values here, including `selected`, which I added on judgement before checking. `typography/error` also exists in Figma at size 12, weight 400, line height 16, so adding it to native was right and the values match.

**Two bugs that predate this batch**, both native-only and neither fixed here:

`size.xxlarge` is 44 on iOS, Android and RN. Figma says 40, and web already uses 40. It is live: Android uses it as the minimum button height. So native buttons and fields are 4px taller than design specifies.

The dark-mode greys do not match Figma on six of seven tokens. Figma and web write them as a translucent light over the sheet; native pre-composites them to opaque values. In light mode the two are identical, which is why this went unnoticed. In dark mode they diverge by up to 45/255. `gray.300`, the default input border, renders at 53 on web against 87 on native.

Raising both separately rather than widening this PR.

bthebladeprimer (approved):

OnurVar:

PR #1882, opened 2026-08-31
fix(components): let the co-badge chips inherit the field (ORC-8267)

OnurVar:
# Description

[ORC-8267](https://primerapi.atlassian.net/browse/ORC-8267)

The co-badge network chips painted the sheet colour behind themselves, so once a merchant coloured the field differently they showed as pale rectangles on top of it. They're transparent now and inherit the field; selection reads from the border alone, as in Figma and on Android.

They no longer need a `tokens` parameter, which is why the diff is mostly deletion.

# Contributor Checklist

- [x] I have manually tested newly added UX

[ORC-8267]: https://primerapi.atlassian.net/browse/ORC-8267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bthebladeprimer (approved):
Approved but would be nice to fix SonarCloud check failures

PR #1890, opened 2026-09-02
feat: Open a CheckoutComponents demo straight from a deep link (ORC-7710)

bthebladeprimer:
# Description

ORC-7710

Lets a deep link open a CheckoutComponents demo directly, so an E2E scenario reaches a CC surface with no taps. The Debug App already accepted `primer://sdk-demo.primer.io/…?clientToken=…&settings=…` for Drop-in and Headless; CheckoutComponents was unreachable in that mode and the link could not select anything.

- Add an optional `demo=<key>` query parameter. With it the settings screen applies the token and settings as before and then presents the demo itself. Without it the link behaves as today, so the drop-in E2E suite and the Appetize links are unaffected.
- Give every demo a stable key (`DemoKey`, 30 cases). The raw values are a cross-platform contract shared byte for byte with Primer Studio (Android) and the E2E page objects, so they are literals, never derived from a display name. `DemoMetadata` is identified by the key instead of a per-launch UUID, and each Examples row exposes `demo_row_<key>`.
- Fix a pre-existing cold-start hole: the URL handler posted its notification before the settings screen registered its observer, so a link that launched the app was dropped, token and all. The handler now keeps the payload as a pending inbox that `viewDidLoad` consumes. Presentation waits for `viewDidAppear`, pops any pushed screen and dismisses any modal first.
- Honour the deep-linked settings on the CheckoutComponents path (it always rebuilt them from the form) and stop hiding the CheckoutComponents button in AppLink mode.
- Surface failures: a malformed link shows `deep_link_error_label` with `deep link error: <reason>` instead of a silent no-op. The AppLink "clear" button drops its duplicated `Primer SDK Button` identifier, which the drop-in page object queries.
- Pin the keys in `DemoKeyContractTests` (runs in the CI `DebugAppTests` scheme), cover the parser and the pending inbox in `AppLinkConfigProviderTests`, and add a Danger check that every key backs exactly one registered demo. `Debug App/README.md` documents the link.

No SDK source changes. The shared key table lives next to the CC scenario matrix in the e2e-tests repo ([MR !588](https://gitlab.com/primer-io/processing/e2e-tests/-/merge_requests/588)).

# Other Notes

- The SwiftFormat pre-commit hook reformatted the whole `Dangerfile.swift` and the header year of `AppLinkConfigProviderTests.swift`, so those files carry a little unrelated churn. The Dangerfile block will need a trivial merge with the liveness check from #1886.
- `MerchantSessionAndSettingsViewController.swift` already exceeds the SwiftLint file-length limit on the base branch; this PR adds about 24 lines to it and keeps the new logic in a separate extension file.

# Manual Testing

Verified live with a local Appium session (XCUITest, `mobile: deepLink` with `bundleId`, the same delivery the wdio suite uses) on an iPhone 16 simulator:

- Link to a running app with `demo=default_checkout`: the payment-method list is findable about 2 s after the call. `demo=inline_card_form`: the card number field is findable about 2.6 s after the call.
- Link that launches the app (cold start, both `mobile: deepLink` and `xcrun simctl openurl`): the demo is presented; first CC element about 5 s after the call.
- Link while another demo sheet is up, and while the drop-in screen is pushed on top: the current screen is replaced by the requested demo.
- Cancel and the demo's Done alert dismiss the presented demo.
- A link with an unknown key shows `deep link error: unknown demo key "nope"`; a link without settings shows `deep link error: settings missing`; a valid link clears the label. A 2-parameter link lands in AppLink mode with the CheckoutComponents button visible and no error.
- `~Primer SDK Button` now resolves to exactly one element in AppLink mode.
- Full card payment on `default_checkout` through the new entry: success screen shown for its 3 s window, sheet dismissed after.

# Contributor Checklist

- [x]  I have open a documentation PR, if applicable

# Reviewer Checklist

# Before Merging


Debug App/Sources/Utilities/AppLinkConfigProvider.swift
```diff
@@ -37,6 +43,12 @@ class AppLinkConfigProvider {
         return settings
     }
 
+    /// Why `fetchConfig()` returns nil, worded for the settings screen's error label; nil when it decodes.
+    var settingsFailure: String? {
```
henry-cooper-primer: above init

Debug App/Sources/Utilities/AppLinkConfigProvider.swift
```diff
@@ -58,19 +70,29 @@ extension UserDefaults: AppLinkPayloadProviding {
...
-        let settings = URLComponents(url: url, resolvingAgainstBaseURL: true)?.queryItems?.first(where: { $0.name == "settings"}) {
-            let DeeplinkConfigProvider = DeeplinkConfigProvider(clientToken: clientToken.value, settingsJwt: settings.value)
-            NotificationCenter.default.post(name: .appetizeURLHandled, object: DeeplinkConfigProvider)
-            return true
-        } else {
-            return false
-        }
+        guard url.host == host else { return false }
+        let queryItems = URLComponents(url: url, resolvingAgainstBaseURL: true)?.queryItems ?? []
+        func value(_ name: String) -> String? { queryItems.first { $0.name == name }?.value }
```
henry-cooper-primer: quite unusual having a function within a function

Debug App/Sources/Utilities/AppLinkConfigProvider.swift
bthebladeprimer: resolved

Debug App/Sources/Utilities/AppLinkConfigProvider.swift
bthebladeprimer: resolved

nquinn-primer (approved):

PR #1893, opened 2026-09-03
fix(components): drop the duplicate inputText colour helper

OnurVar:
# Description

`CheckoutColors` declares `inputText(tokens:)` twice, so the branch does not compile.

Neither PR did this. #1877 added the helper and merged with one copy. #1878 was stacked on #1877, so its branch carried the helper too, and once #1877 had been squash-merged its commits no longer matched anything here, so the squash of #1878 was computed against an older base and applied the addition a second time. `ov/fix/ORC-8179-label-on-brand` itself has exactly one copy.

Deleting the second one. The file is byte-identical to #1878's branch afterwards.

# Other Notes

Worth switching the rest of the theming stack to merge commits rather than squashes. Every remaining PR sits on the one below it, so the same thing can happen again, and this time it landed with no conflict to warn anyone.

# Contributor Checklist

- [x]  I have added UI tests to new user flows, if applicable
- [x]  I have open a documentation PR, if applicable

# Reviewer Checklist


OnurVar:
Landing the four-line deletion directly on the integration branch instead.

PR #1896, opened 2026-09-07
fix(components): repaint payment method tiles on a colour-scheme change (ORC-8363)

OnurVar:
# Description

ORC-8363

Open the checkout in dark mode and switch the phone to light while the sheet is open. The payment method tiles kept their dark background, border, border width and logo while the labels switched. Same the other way round.

The tile colours came out of `BaseColors.uiColor` as a fixed colour for whichever scheme was active when the list loaded, and the list is built once. `uiColor` now returns a dynamic colour that resolves against the trait collection it is drawn in, so `Color(uiColor)` repaints on a scheme change with no call-site change. Border width and logo cannot be dynamic values, so the metadata variant objects now travel with the payment method (internal fields; the public `icon` and `borderWidth` snapshot stay as they were) and `PaymentMethodButton` picks the variant from the environment colour scheme, the same source the design tokens use.

A merchant who forces light or dark through `appearanceMode` is unaffected. The sheet already pins its traits (`overrideUserInterfaceStyle` on the bridge controller, `preferredColorScheme` in `PrimerCheckout`), so trait-based resolution lands on the forced scheme.

No breaking changes. The public surface of `CheckoutPaymentMethod` is unchanged.

# Other Notes

- Drop-in is untouched. `PrimerPaymentMethod.logo` now goes through `BaseImage.image(isDark:)` with the same fallback order; `invertedLogo` and `UIScreen.isDarkModeEnabled` are as before.
- `hasVisibleBackground` resolves the colour for the current scheme before reading its components. `getRed` on a dynamic colour would otherwise use `UITraitCollection.current`.
- A malformed hex in the metadata keeps the old one-shot resolution for that colour instead of leaking the other scheme's value. Valid metadata is what gets the dynamic colour.
- Merchants rendering their own tile through the `PrimerPaymentMethods` `method` slot get the dynamic colours for free, but `icon` is still the load-time snapshot. A scheme-aware image accessor would be a public API addition, left for a follow-up.
- Origin: listed in the ORC-8178 findings on 2026-08-20. #1884 fixed the bridged input fields only, which had a different cause.
- `CheckoutComponentsPaymentMethodsBridgeTests` now registers the three payment method types it uses in `setUp`. Before, the class only passed when earlier test classes had filled the registry, and crashed with an index-out-of-range when run on its own.

# Manual Testing

Debug App, SPM target, Checkout Components, default checkout demo. Reproduced on the pre-fix build first, then verified on the fix; all four scenarios pass.

1. Simulator in dark, open the sheet, let the list render.
2. Cmd+Shift+A to light without leaving the sheet. Tiles repaint background, border and logo.
3. Repeat starting from light.
4. `appearanceMode` forced to `.light`, then `.dark`: tiles ignore the device switch.

Unit tests: `PrimerThemeImagesTests` (new), `InternalPaymentMethodTests`, `CheckoutComponentsPaymentMethodsBridgeTests`, `HeadlessRepositoryGetPaymentMethodsTests`, `PaymentMethodMapperTests`; `PrimerPaymentMethodTests` still covers `logo`.

# Contributor Checklist

# Reviewer Checklist

# Before Merging

# Other Stuff

### DIFF
--- a/Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Bridge/CheckoutComponentsPaymentMethodsBridge.swift
+++ b/Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Bridge/CheckoutComponentsPaymentMethodsBridge.swift
@@ -77,7 +77,9 @@ final class CheckoutComponentsPaymentMethodsBridge: GetPaymentMethodsInteractor,
         textColor: textColor,
         borderColor: borderColor,
         borderWidth: borderWidth,
-        cornerRadius: cornerRadius
+        cornerRadius: cornerRadius,
+        logoVariants: primerMethod.baseLogoImage,
+        borderWidthVariants: displayButton?.borderWidth
       )
     }
 
--- a/Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Data/Mappers/PaymentMethodMapper.swift
+++ b/Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Data/Mappers/PaymentMethodMapper.swift
@@ -27,7 +27,7 @@ final class PaymentMethodMapperImpl: PaymentMethodMapper {
     let formattedSurcharge = formatSurcharge(
       internalMethod.surcharge, hasUnknownSurcharge: internalMethod.hasUnknownSurcharge)
 
-    return CheckoutPaymentMethod(
+    var method = CheckoutPaymentMethod(
       id: internalMethod.id,
       type: internalMethod.type,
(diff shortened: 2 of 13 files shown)


OnurVar:
The unit-test job here is the ORC-8364 crash/hang (no failing test in the log, one counted failure after the bundle restarts), fixed by #1895. This PR touches none of that code. Once #1895 lands on bn/feature/checkout-components I'll update the branch and re-run.

bthebladeprimer (approved):

PR #1900, opened 2026-09-11
refactor(components): make the raw palette colours internal (ORC-8396)

OnurVar:
# Description

[ORC-8396](https://primerapi.atlassian.net/browse/ORC-8396)

The theme mixed two kinds of colour: ones named after their job (`primerColorBorderOutlinedDefault`) and ones named after the paint (`primerColorGray300`). A merchant setting the second kind has no way to know what it moves, and we were free to change it in any release without that counting as breaking.

This removes the 14 palette colours from `ColorOverrides`: the eight greys, `primerColorGreen500`, the three reds and the two blues. They stay in the token file and are still what the purposeful colours resolve from, so nothing looks different.

`primerColorBrand` stays, because it says what it is for. It is still injected before references resolve, so `borderOutlinedSelected`, `focus` and `loader` follow it.

Breaking: a merchant setting any of those 14 no longer compiles. They set the purposeful colour instead, which is what the other SDKs are moving to in the same ticket.

# Other Notes

Four tests covered the injection cascade using `gray300`. Rather than delete them they now use `primerColorBrand`, so the mechanism stays covered. Two tests that only set greys and asserted they took have gone.

Web is deliberately not part of this ticket. Its purposeful tokens are defined as the palette (`background.primary` is `var(--primer-color-gray-000)`), so the palette has to stay in the stylesheet and any merchant CSS can reach it.

# Manual Testing

None needed, no visual change.

# Contributor Checklist

- [x]  All status checks have passed prior to code review

# Reviewer Checklist

# Before Merging

[ORC-8396]: https://primerapi.atlassian.net/browse/ORC-8396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Tokens/DesignTokensManager.swift
+++ b/Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Tokens/DesignTokensManager.swift
@@ -37,7 +37,7 @@ final class DesignTokensManager: ObservableObject {
     tokens = loadedTokens
   }
 
-  /// Injected before references resolve, so tokens aliasing a palette entry or the brand font follow the override.
+  /// Injected before references resolve, so tokens aliasing the brand colour or the brand font follow the override.
   private func tokenValueOverrides(colors: ColorOverrides?) -> [String: Any] {
     var overrides: [String: Any] = [:]
     if let brandFont = themeOverrides?.typography?.brand {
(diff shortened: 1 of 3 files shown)


Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Tokens/DesignTokensManager.swift
```diff
@@ -37,7 +37,7 @@ final class DesignTokensManager: ObservableObject {
     tokens = loadedTokens
   }
 
-  /// Injected before references resolve, so tokens aliasing a palette entry or the brand font follow the override.
+  /// Injected before references resolve, so tokens aliasing the brand colour or the brand font follow the override.
```
henry-cooper-primer, line 40: let's stick to US spelling in code and comments

henry-cooper-primer (approved):

Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Tokens/DesignTokensManager.swift
OnurVar: Done. Swept all twelve in both files, not just this line, since they were all mine.

PR #1904, opened 2026-09-15
refactor(components): put every primary button on a shared PrimerCheckoutButton (ORC-8435)

OnurVar:
# Description

[ORC-8435](https://primerapi.atlassian.net/browse/ORC-8435)

Android, React Native and web each have one button component, so a merchant moving a token moves every button together. iOS had eleven written inline in five different shapes. Corner radius was small on six and medium on five, the label was one of three styles, the height was built three ways and landed on four values, and only two of the eleven set a height at all.

`PrimerButton` is that one component, measured from the design: 44 minimum height, medium radius, `titleLarge` label, filled and outlined as two styles of one thing. Same name as the other three SDKs.

Every brand-filled button now goes through it: the card form's Pay, the vaulted Pay in both places, ACH continue and accept, Klarna's two, the Apple Pay fallback, the billing-address redirect, the form redirect, the delete confirm and the error retry.

## What visibly changes

Six buttons that were on the small radius move to 8. The label goes to `titleLarge` on the ones that were on `body` or `bodyMedium`. Buttons that set no height, which was nine of them, now have a 44 floor. The Apple Pay fallback's bare `height: 50` is gone.

The loading rule is in one place now: a loading button keeps its own resting look and adds a spinner, so a shopper never sees the disabled grey on a button that is taking their money. Each screen used to spell that out itself, and the delete confirm did not, it just kept the brand fill with no enabled/disabled distinction.

## Two new colour helpers

`CheckoutColors` had no `textDisabled` or `borderDisabled`. The outlined style needs both, so they are added next to the ones already there, reading the tokens that already exist.

`PrimerComponentHeight.interactive` is the shared 44.

## Not in this PR

The outlined variant is built and unused so far. The four bypassing field families, the field container's disabled state and the Cancel control are the second iOS PR, as planned.

# Other Notes

**Not compiled locally.** The checkout Xcode has open is behind this branch and carries another session's uncommitted test files, so a fast-forward would land on their work. Every changed file parses clean, SwiftFormat and SwiftLint are clean, and CI is the first real build.

Part of a parity batch across the four SDKs.

# Manual Testing

Walk the sheet and check every button is the same shape, the same height and the same label weight. Start a card payment and a vaulted payment and watch the button stay brand-filled with a spinner rather than greying out. Delete a saved card and check the confirm button spins.

# Contributor Checklist

- [ ] I have added unit tests to a reasonable level of coverage where suitable

[ORC-8435]: https://primerapi.atlassian.net/browse/ORC-8435?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bthebladeprimer (approved):
Amazing!

PR #1906, opened 2026-09-15
feat(components): offer autofill on the card and billing fields (ORC-8442)

OnurVar:
# Description

[ORC-8442](https://primerapi.atlassian.net/browse/ORC-8442)

A shopper with a saved card gets a one tap fill on the React Native checkout and types sixteen digits by hand on iOS. The twelve card and billing fields offered nothing, so the OS showed no saved card, no camera scan and no saved address. Only the one time code fields and the MB WAY phone set anything.

Every field now names what it holds: card number, cardholder, first and last name, both address lines, city, state and postal code.

The billing phone also moves off an alphabetic keyboard with word capitalisation onto a phone pad. A phone configuration existed and nothing used it.

# Other Notes

The fields that differ only in what the OS should offer share one preset through a small `offering(_:)` helper, rather than gaining a preset each. The two views that serve several fields, name and address line, pick from the element type.

Expiry and CVV stay without a content type on purpose.

Not compiled locally, same Xcode reason as the other iOS PRs in this batch. Every changed file parses clean, SwiftFormat and SwiftLint clean.

Part of a parity batch. Android, React Native and web have their own PRs.

# Manual Testing

Needs real saved cards and contacts, not typed input. Open the card form with a card saved in Safari or iCloud Keychain and check the keyboard offers it, and that the camera scan appears on the card number. Fill a billing address from a saved contact.

# Contributor Checklist

- [ ] I have added unit tests to a reasonable level of coverage where suitable

[ORC-8442]: https://primerapi.atlassian.net/browse/ORC-8442?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bthebladeprimer (approved):
Great job!

PR #1907, opened 2026-09-15
fix(components): lock the fields, share the field container and Cancel (ORC-8435)

OnurVar:
# Description

ORC-8435

ORC-8435's second iOS half: the field container's locked state, the four field families that bypassed that container, and Cancel.

**Locked state.** `backgroundOutlinedDisabled` is a merchant setting on all four SDKs and reached a field on Android only. The iOS container had no disabled state at all, so fields looked exactly as they had a moment earlier while a payment ran. The container now takes the disabled fill, border and label colour, matching Android !429 and React Native #436.

Done with a new `\.isInputEnabled` environment value rather than a parameter. Thirteen views build a container and none of them knew the form was busy, so threading a flag would have been thirteen signature changes for one fact the form already had. `CardFormFieldsView` sets it once from `cardFormState.isLoading`. Same shape as Android's composition local, and it defaults to true so nothing outside the card form moves.

**The four bypassing families.** All four drew their own box, so all four dropped the error icon, two had no focus state and two used the wrong label colour.

- The billing redirect's text field and its country menu, and the form redirect's field, now use `PrimerInputFieldContainer`. They get the error icon, the focus border, the error border width and the container's label colour. The billing screen's per-field focus comes from one screen-level `@FocusState`; the form redirect field mirrors its `@FocusState` into the container's binding.
- The vaulted CVV keeps its own box. It sits inline beside the instruction text and is 120pt wide, so the container's label-above / error-below stack does not fit it. It takes the two things it was actually missing: the error icon, and a border that widens on error rather than only on focus. The icon itself is now one `PrimerFieldErrorIcon` that the container uses too.

The form redirect field loses nothing: its helper text still renders below the container when there is no error.

**Cancel.** One colour on seven screens, another on four, a third on one. It is now `CheckoutHeaderButton` on all twelve, which is the header's own right-hand control lifted out of `CheckoutHeaderView` so the screens that lay out their own header row can use it. A merchant moving `textPrimary` now moves every Cancel.

The one exception is the delete-confirmation screen, where Cancel is a full-width outlined button, not a header link. It is now `PrimerCheckoutButton(style: .outlined)`, which is the outlined variant #1904 built and had no caller for.

# Other Notes

Visible changes to expect on a device:

- Cancel is `titleLarge` and `textPrimary` everywhere. It was smaller and grey on seven screens, and blue in the country picker's nav bar.
- The three migrated field families move from padding-based height to the container's fixed height, so they get slightly shorter and line up with the card form fields. ORC-8444 raises all of them together.
- The billing country menu's chevron now sits next to the error icon when that field is in error.

Based on #1904, not the integration branch. They share `textDisabled` and `borderDisabled` on `CheckoutColors`, the outlined button style, and ORC-8435 orders them this way.

# Manual Testing

Not compiled locally. The Xcode checkout is still parked behind two uncommitted test files from another session, so CI is the first build. Parses clean, SwiftFormat and SwiftLint clean.

This one wants a device pass more than the others in the batch: three payment screens change layout.

# Contributor Checklist


Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Presentation/Components/CheckoutHeaderButton.swift
```diff
@@ -0,0 +1,40 @@
...
+@_spi(PrimerInternal) import PrimerFoundation
+@_spi(PrimerInternal) import PrimerCore
+
+/// The trailing control in a checkout header: Cancel, Edit or Done.
+///
+/// `CheckoutHeaderView` draws it, and so do the screens that lay out their own header row, so one
+/// token moves every one of them. Cancel used to be three different colours across twelve screens.
+@available(iOS 15.0, *)
+struct CheckoutHeaderButton: View {
+  let config: CheckoutHeaderView.RightButtonConfig
```
henry-cooper-primer: private

Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Presentation/Components/VaultedCardCVVInput.swift
```diff
@@ -127,6 +128,16 @@ struct VaultedCardCVVInput: View {
 
   // MARK: - Helpers
 
+  private var cvvBorderWidth: CGFloat {
```
henry-cooper-primer: above funcs

Sources/PrimerSDK/Classes/CheckoutComponents/Internal/Presentation/Screens/BillingAddressRedirect/BillingAddressRedirectScreen.swift
```diff
@@ -195,46 +184,31 @@ struct BillingAddressRedirectScreen: View {
...
-      : CheckoutColors.borderDefault(tokens: tokens)
+  private func errorBinding(for fieldType: PrimerInputElementType) -> Binding<String?> {
+    .constant(billingState.errors[fieldType]?.message)
   }
 
-  private func fieldBorderWidth(for fieldType: PrimerInputElementType) -> CGFloat {
-    billingState.errors[fieldType] != nil
-      ? PrimerBorderWidth.error(tokens: tokens)
-      : PrimerBorderWidth.standard(tokens: tokens)
+  /// The container only reads this, and focus is owned by `focusedField`.
```
henry-cooper-primer: superflous comment

bthebladeprimer (approved):

PR #1909, opened 2026-09-16
fix: don't set a delegate on FileManager.default — crashes on iOS 27 (ORC-8465)

OnurVar:
# Description

[ORC-8465](https://primerapi.atlassian.net/browse/ORC-8465)

On iOS 27, opening a checkout crashes in `-[NSFileManager setDelegate:]`. `Downloader` set a delegate on `FileManager.default` before copying downloaded images, and iOS 27 now traps on that. It affects every merchant on iOS 27, with the released SDK too.

`Downloader` now uses its own `FileManager` instance for the copy. No public API changes, no breaking changes.

# Manual Testing

Debug App on an iOS 27 simulator, clean install: Universal Checkout crashed before, renders after. Also verified on the checkout-components branch and through the React Native example.

Note: the crash only reproduces on a fresh install — an app that already has its images cached skips the failing path.

[ORC-8465]: https://primerapi.atlassian.net/browse/ORC-8465?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bthebladeprimer (approved):

OnurVar:
Parked: the iOS 27 fixes go to the checkout-components branch for now — see #1914. Branch kept.

PR #1912, opened 2026-09-17
feat(components)!: Close the CheckoutComponents gaps the account funding demo exposed (ORC-8477)

bthebladeprimer:
# Description

ORC-8389 and ORC-8477.

Two halves. The demo rebuilds a beta merchant's deposit flow on the inline SwiftUI integration, and in doing so it pinned down a set of CheckoutComponents gaps. The second half closes them. Every gap was checked against Android v3 first, because Android is the source of truth for the CheckoutComponents API; in each case the fix copies Android's shape rather than inventing an iOS one.

## SDK changes (ORC-8477)

- **Paying with a saved method.** `PrimerSelectionSession.selectVaulted(_:)` now charges the method instead of only marking it, matching Android's `PrimerVaultedPaymentMethodsController.select(method)`. The internal `submitSelectedVaulted()` is deleted. A merchant with their own list and their own pay button previously had no way to pay at all.
- **CVV recapture moves to its own screen.** New `CheckoutNavigationState.cvvRecapture` plus `VaultedCardCvvRecaptureScreen`, mirroring Android's `Screen.CvvRecapture` and Drop-In's `CVVRecaptureViewController`. The feature is unchanged: same client-session gate, same validation, same `PrimerVaultedCardAdditionalData` on the same exchange call. Only the field's home moves, so a merchant-owned pay button can no longer stall on a card that needs a code.
- **The processing screen shows on the saved-card path.** `executeVaultPayment()` never called `startProcessing()`, so the whole payment ran with nothing on screen. Android's `submitVaultedPaymentInternal()` does both.
- **The saved-method list is published.** It was a plain computed getter, so a merchant's own list never re-rendered after `delete(_:)`. Android's equivalent is a `StateFlow`.
- **`formatAmount(_:)` on `PrimerCheckoutSession`.** Mirrors Android's `PrimerCheckoutController.formatAmount(Int)`. Decimal digits come from the client session's currency and cannot be worked out from the currency code alone.
- **The session modifier takes a theme.** `.primerCheckoutSession(_:theme:onCompletion:)` re-themes on every change, as Android's `PrimerCheckoutHost` does.
- **`API_REFERENCE.md` corrected** declaration by declaration, along with `README.md`, both `CLAUDE.md` files and the context rules.

BREAKING CHANGE: `PrimerSelectionSession.selectVaulted(_:)` keeps its signature and changes meaning, from "highlight this saved method" to "charge this saved method". Existing code that calls it from a row tap will compile and then take money. Move the highlight into your own view state and call `selectVaulted(_:)` only from your pay button.

BREAKING CHANGE: two published merchant-docs pages carry iOS samples that call `selectVaulted(_:)` from a row tap, so under this change they charge the customer on tap. `display-saved-payment-methods.mdx` and `events.mdx` are rewritten in primer-io/general/public-docs!376, which should land with this.

BREAKING CHANGE: the inline vaulted CVV surface is removed. `PrimerPaymentMethodSelectionState.requiresCvvInput`, `.cvvInput`, `.isCvvValid` and `.cvvError`, `PrimerSelectionSession.updateCvvInput(_:)` and `VaultedPaymentMethodsDefaults.cvvInput(_:)` no longer exist. CVV recapture still happens; the SDK now raises its own screen for it, so nothing replaces them.

## Follow-up fixes in the same PR

Reviewing the four items above against the source turned up four defects, two of them introduced by this PR. All are fixed here rather than deferred.

- **Every amount now formats in the configured locale.** `formatAmount` and the selection header honoured `PrimerSettings.localeData`; the card form's pay button, the QR amount and all three surcharge strings still used the device locale, so a merchant's own total agreed with one SDK screen and disagreed with another. All six sites now take the locale from `ConfigurationService` alongside the currency they already take from it. Merchant-visible. The derived locale yields the device locale untouched when the configured code already describes the device, because `PrimerLocaleData` carries only a language and a region and rebuilding from it would turn `sr-Latn-RS` into `sr-RS` for everyone who never configured one.
- **The CVV screen's pay button gets its own identifier.** Moving recapture onto its own screen left `Vault.payButton` live there and on the selection screen's vault section at the same time, which makes an automation selector ambiguous. The new screen takes `Vault.cvvPayButton`; `Vault.payButton` keeps its string and stays where it was. Recorded in primer-io/processing/e2e-tests!597.
- **Retry re-runs the payment that actually failed.** Pre-existing, and this PR widens who reaches it. The error screen read its target from the scope of the last payment-method screen opened, which is never cleared on the way back to selection. Open the card form, go back, pay with a saved card, have it decline, tap Retry: it submitted the card form. With a filled form that tokenizes and charges a different card than the one being retried. A stale PayPal or Klarna scope relaunched that method. It also emitted a reattempt event with nothing to re-run. What to re-run is now recorded where a payment starts, taken from the caller rather than the navigation state, which stays on selection for the whole of an inline flow. A saved-card retry goes through the full vault entry point, so a card needing CVV asks for it again and no code is held over.
- **Analytics report the integrated product, not the routing surface.** See item 2 below.

## The demo (ORC-8389)

A wallet top-up flow in a made-up merchant brand. The merchant owns the amount screen, the saved-card list, the add-a-card panel, the pinned pay bar and both result dialogs. Primer supplies the vault, the card fields and the payment.

- `PrimerSelectionSession.vaultedPaymentMethods` iterated into the merchant's own list, collapsed to the highlighted card with a "View all" toggle, the highlight held in the demo's own state and the pay bar calling `selectVaulted(_:)`
- `PrimerPaymentMethods`'s `method` slot styled as the merchant's brand buttons, with the card row opening the merchant's panel instead of the SDK flow
- `PrimerCardForm` recomposed from `CardFormDefaults.cardholderName / cardNumber / expiryDate / cvv / postalCode` inside a merchant bottom panel with its own Cancel and Confirm
- `isSuccessScreenEnabled` and `isErrorScreenEnabled` off, so the merchant's own dialogs render from `onCompletion`, with the returned payment id printed on the success dialog

# Other Notes

All eight items are implemented, and the four notes that stood here are now resolved rather than deferred.

1. **Item 8 is verified.** Captured on the simulator during a CheckoutComponents card payment that reached 3DS: Netcetera's branded progress dialog renders for the full EMVCo window, unobscured, over a dimmed checkout. Ten frames were captured; they go on ORC-8477. Two corrections to what stood here: `showProgressDialog()` does gate window z-order on `.dropIn` only, and CheckoutComponents itself creates no window, but `3DSService.perform3DSChallenge` creates a Primer-owned `UIWindow` at `.normal` for every integration type while the dialog is deliberately still up. That is shared code, identical under Headless, so the risk comparison stands but the absolute wording did not. At 3DS time CheckoutComponents is already `.headless`, so it runs the branch commented "Headless mode" rather than falling through a distinct value.
2. **`sdkIntegrationType`, fixed here.** The overwrite is real and arrives from three paths, not one: `RawDataManager.init`, `VaultManager.init` and the `PrimerHeadlessUniversalCheckout.delegate` didSet, the last of which runs first in the CheckoutComponents flow. Correcting the value itself would break payments, because `PrimerDelegate` routes tokenization and resume callbacks to CheckoutComponents' own completion handler only under the `.headless` branch. So the two jobs the value does are separated instead. `sdkIntegrationType` keeps routing, untouched, and `sdkIntegrationProduct` is written only where a checkout starts and read by the two analytics call sites with the routing value as a fallback. No payment behaviour changes on any integration, and `preferredNetwork(from:)` and `PrimerDelegate` are not touched.
3. **`formatAmount`.** The iOS behaviour is unchanged and now consistent, see the locale fix above. The two claims about Android are not verifiable from this repo, so treat them as carried from the earlier parity study rather than as checked here. Worth a confirmation from Darius.
4. **`retryPayment()` was not a no-op**, and the reason given was wrong. Fixed here, see above.

One unrelated log message is also corrected. `3DSService` quoted `https://applink` inside a sentence saying the value was not in a valid format, so it read as a value the merchant had set. It is a fixed example, and the warning fires when nothing usable is configured at all. Log text only, no behaviour change, and it cost real time to work out during this ticket.

Docs outside this repo are open as merge requests rather than left to follow:

- primer-io/general/public-docs!376, seven pages. Four carried substantive stale content, two of them the working iOS samples named in the breaking change above.
- primer-io/processing/e2e-tests!597, the identifier convention and the `account_funding` deep-link row.

## Recordings

Both recorded on iPhone 17 Pro / iOS 26.0 against Sandbox, with two saved cards on the customer and `Enable CVV Recapture flow` on.

https://github.com/user-attachments/assets/98b58d59-61bc-4028-a179-d7fbba530bc0

**Saved card with CVV recapture.** Tapping a saved-card row moves the tick and charges nothing, which is the item 1 regression. The pay bar then charges the highlighted card, and the SDK raises its own Enter CVV screen for it, which is item 2. Validation holds the Pay button until the third digit. Submitting shows the SDK's processing screen, which is item 3 and did not exist on this path before.

https://github.com/user-attachments/assets/d12a45d6-25fa-45f9-8468-aae163635954

**Saved card without recapture.** The same pay bar goes straight to the processing screen and on to the merchant's own result dialog, which prints the payment id Primer returned.

Everything above the SDK's screens is merchant code: the list, the highlight, the pay bar and both dialogs.

# Manual Testing

Full unit test plan green, 0 failures. SwiftFormat and SwiftLint clean on every changed file. Both the `PrimerSDK` and `Debug App` schemes build.

Walked end to end on the simulator (iPhone 17 Pro, iOS 26.0) against Sandbox, with a card vaulted through the Drop-In vault manager first and `Enable CVV Recapture flow` on:

1. Account Funding demo opens, amount screen creates the client session, payments screen renders the merchant's own list with `Visa **** 4242` and the pinned pay bar showing the client-session amount.
2. **Tapping the saved-card row highlights it and does not charge.** This is the item 1 regression that would otherwise take money from a row tap.
3. **The pay bar raises the SDK's own "Enter CVV" screen**, showing the card being paid for, the instruction line and a focused field. That is item 2: a merchant-owned pay button on a card that needs recapture used to have nowhere to put the field.
4. **Validation gates the button.** Two digits leaves Pay greyed out, the third enables it.
5. **Submitting shows the SDK's processing screen**, which is item 3. The saved-card path previously ran the whole payment with nothing on screen.
6. The payment completes as a real Sandbox payment and the merchant's own success dialog renders with the returned id, `Payment JmrvxJv9G`.

Also confirmed: with no saved method the list correctly renders nothing and the pay bar stays disabled, and the saved card appears without restarting the session, which is the published list from item 4.

A second pass covered the follow-up fixes, on the Default Checkout demo:

7. **Item 8.** A card payment reached 3DS and Netcetera's progress dialog rendered unobscured for the full EMVCo window, captured frame by frame across ten frames.
8. **Retry on the saved-card path.** Paid with the saved card, it declined, tapped Retry. The logs show a second, distinct vault payment through `executeVaultPayment` with its own payment id, rather than the previous silent no-op.
9. **The locale fix.** The card form's pay button and the selection header both render `6,00€` from the session's configured locale, where the button previously used the device's.

Full unit plan run repeatedly. One failure kept surfacing in `test_navigationStateStream_consumerEarlyExit_terminatesCleanly`, a test this PR does not touch, with the container reported as terminated. I first read it as the known ORC-8484 flake, which was wrong. It was mine: both new test classes write into `DIContainer.shared` without resetting it, which the repo's own `ContainerTestHelpers` says to do on both sides. Their scopes' async init then raced whichever class cleared the container next, and the failure landed somewhere else. Fixed in the last commit. The underlying fragility that lets one class's leak surface in another is still ORC-8484.

Not covered:

- the no-CVV vaulted path through the funding demo, because it reaches the same `executeVaultPayment` the saved-card run exercised. It has its own unit test.
- items 5 and 6 at runtime, since the demo uses neither. Both have unit tests.
- CVV recapture after the retry change was covered by unit test rather than by hand, because the card the Sandbox returned did not require it on this pass. The screen itself is unchanged apart from its pay button identifier.

Sandbox itself is unreliable for this app: attempts end in `[failed-to-resume-payment]` on the card path, the vault path and other methods alike, in the same session. Not specific to these changes. The `threeDsAppRequestorUrl` warning that appears alongside is unrelated to those failures: the Debug App sets no value, the warning is non-fatal and only concerns out-of-band redirect-back.


henry-cooper-primer:
this is very large, can it be broken down?

bthebladeprimer:
Closing in favour of three smaller PRs, on Henry's review.

- #1918, two unrelated fixes found on the way. 8 files.
- #1919, the SDK change. 52 files. This is the one that needs the review.
- #1920, the demo. 9 files under `Debug App` only, stacked on #1919.

Same content, split by what it touches. The recordings moved to #1919, where the SDK screens they show are changed.

I closed this rather than shrinking it because the branch behind it was named for the demo, and GitHub cannot repoint an open PR at another branch. The combined branch is kept as `bn/archive/ORC-8477-combined-before-split`, so nothing here is lost.

